PR #6, opened 2022-04-30
fix Archimedean cdf and some typos

mlkrock:
Hi @lrnv thanks for this great package.

I did some basic typo fixes and fixed the Archimedean cdf, which had some issues. I chose to loop through the values in this cdf calculation so `ϕ` and `ϕ⁻¹` didn't have to be written with vector operations like `exp.(u)` or defined to broadcast (currently can't broadcast).


lrnv:
Hey ! Thanks a lot for taking the time to PR this. I did not ran it but seems ok to me, I'll merge. 

Glad you found the package useful :) Still lacks tests and docs though, this will probably improve this summer after my PhD end.

Issue #7, opened 2022-04-30
pdf/logpdf for Archimedean copula


mlkrock:
Looking at R copula library, the calculation of the pdf/logpdf is quite a pain.

If you don't mind the dependency, could use ForwardDiff.jl to get the derivatives of the generator, then everything is pretty easy. The main question is how to get the  higher order derivatives, I see some discussion e.g. https://discourse.julialang.org/t/automatic-differentiation-for-higher-order-derivatives/62337.
https://discourse.julialang.org/t/forwarddiff-3rd-or-even-higher-order-derivatives/8011/6
The easy option is to just write 

`∂ϕ(cop::CT, t::Real) where {CT<:Copulas.ArchimedeanCopula} = ForwardDiff.derivative(x -> Copulas.ϕ(cop,x),t)
∂²ϕ(cop::CT, t::Real) where {CT<:Copulas.ArchimedeanCopula} = ForwardDiff.derivative(x -> ∂ϕ(cop,x), t)
`
etc. There might be a smarter way to do this, but I don't see anything obvious. Since, you only need the d'th and first derivative  of the generator for the pdf, it may be fine for now, but there's probably some way to get the first derivative along the way to the d'th derivative.

The R library also stops after d=6 for Archimedean, so could follow that convention, since at some point Archimedean in high dimension is silly.

lrnv:
What I remember without taking a look at it is that the number of derivatives you need is more or less the dimension of the model. Therefore, my ambition was to use ForwardDiff for the *default* implementation on `ArchimedeanCopula`s, as you point out. Feel free to implement it and PR if you want, otherwise I'll do it this summer. No need to stop arbitrarily at d=6, I plan to just let it be generic.

mlkrock:
I just made a PR but I stopped at d=9, so can already say it's better than R there, but I'm not sure of a way to make it a derivative for  general dimension d.

lrnv:
Hey ! I commented on the PR, let's discuss the implementation details there. 

The Joe problem should maybe become a test case ? I agree with you that this is not satisfactory, as 1/1000 quantiles are sometimes needed when dealing with extremal problems (and therefore should be accurate).

PR #8, opened 2022-05-01
add functionality for pdf/logpdf of Archimedean copulas

mlkrock:
This uses ForwardDiff.derivative to compute pdf/logpdf for Archimedean copulas (issue #7)


lrnv:
Hey ! Thanks a lot for the shot. It seems like you did the right thing, although i still have a few comments : 

1) No need to define 9 different functions: for a given copula, you only need the $d$th derivative right ? In which case, this should be enough and much faster, using `TayrloSeries.jl` instead of `ForwardDiff.jl` :

```{julia}
function ϕ⁽ᵈ⁾(C::ArchimedeanCopula{d},t) where d
    X = Taylor1(eltype(t),d)
    taylor_expansion = ϕ(C,t+X)
    coef = getcoeff(taylor_expansion,d) # gets the dth coef. 
    der = coef * factorial(d) # gets the dth derivative of $\phi$ taken in t. 
    return der
end
```

This will avoid the if / elsif / elsif....: 
```{julia}
    numer = if d == 2
        ϕ⁽²⁾(C, sum_ϕ⁻¹u)
    elseif d == 3
        ϕ⁽³⁾(C, sum_ϕ⁻¹u)
    elseif d == 4
        ϕ⁽⁴⁾(C, sum_ϕ⁻¹u)
    elseif d == 5
        ϕ⁽⁵⁾(C, sum_ϕ⁻¹u)
    elseif d == 6
        ϕ⁽⁶⁾(C, sum_ϕ⁻¹u)
    elseif d == 7
        ϕ⁽⁷⁾(C, sum_ϕ⁻¹u)
    elseif d == 8
        ϕ⁽⁸⁾(C, sum_ϕ⁻¹u)
    elseif d == 9
        ϕ⁽⁹⁾(C, sum_ϕ⁻¹u)
    end
```

can just become : 

```{julia}
numer = ϕ⁽ᵈ⁾(C, sum_ϕ⁻¹u)
```

Which is much cleaner. 


2) Remember that the dimension of the copula is a type parameter, and therefore your use of `d = length(C)` is not needed: use parametric dispatch instead (as is done in the upper code). 

3) Why are the `u::Matrix` versions of `Distributions.cdf/pdf/logpdf` needed ? I thought that we should implement it only if it was faster than letting `Distributions.jl` handle them, which is clearly not the case since you simply loop through. You may remove them and/or construct a relevant test case. 

Appart from these small notes, this seems already pretty good ! Tell me what you think.

lrnv:
Ok it now seems to work correctly. I would still like to add some testing of this functionality, but since ht epackage is completely un-tested yet, this will be done later with the testing of other features. 

Thanks a lot @mlkrock  for providing the implementation !

PR #20, opened 2022-11-22
[New copulas] Fréchet-Hoeffding mixture family

### LINKED ISSUE #14: [New copulas] Fréchet-Hoeffding mixture family
So that W and M are in the same parametric family, ref https://github.com/AnderGray/ProbabilityBoundsAnalysis.jl/issues/42

lrnv:
Adding a mother class for W and M that is a parametric mixture class to contain the Frechet-hoeffding bounds

Should resolve #14.

This is just a first draft, maybe we should use more directly the mixture capacities of `Distributions.jl`


lrnv:
Closing this 2022 draft together with #14. The mixture design remained unresolved, the prototype has no meaningful test coverage, and rebasing it would cost more than a fresh implementation if a concrete use case emerges.

PR #37, opened 2023-10-24
Remove precompilation and AlphaStableDistributions.jl dependency

lrnv:
This PR removes the dependency on alphastabledistributions.jl which is causing a lot of load time and compile time to occur,, while we really do not need 99% of it but only one function. 

Sadly, copying the function here does not feel like the right thing to do since it's deduplicating code so I dont know yet what will be the outcome of this.


lrnv:
@codecov-commenter help

PR #38, opened 2023-10-25
[Internals] Switch to TaylorDiff.jl for generator derivatives.

lrnv:
This PR tries to switch from TaylorSeries.jl to TaylorDiff.jl because of better performances of the later. 

But there is still a few broken tests. 

See https://github.com/JuliaDiff/TaylorDiff.jl/issues/61 for upstream report of the bugs. 

See also https://github.com/JuliaDiff/TaylorDiff.jl/issues/62, regarding the amount of additional dependencies, which would need to be merged before we go on here

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - 'lts'
           - '1'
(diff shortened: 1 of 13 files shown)


lrnv:
This makes load time increase significantly... but also reduce compute time for this specific method. Not sure this is worth it.

PR #40, opened 2023-10-26
Bivariate Plackett Copula Addition

Santymax98:
The file PlackettCopula.jl is added that includes: an instance of the bivariate Plackett Copula, calculation of the CDF and PDF functions, the random sample generator and the Spearman Rho calculation. Additionally, I added some testing these functions in biv_plackett.jl.


lrnv:
Hey, 

Thanks a lot for this proposal ! I changed a few things to comply with the current standard, but globally everything is working great. 

**One thing** : Your tests are not passing on the pdf and the cdf, because julia is computing values that are more precise than what you asked for, maybe you should choose is you want to round them or something so that the tests pass ? 

Moreover, where are these values coming from ? the reference ? You could write that as a comment in the test file. 

I am not sure about what to do with the `Base.rand()` and `Distributions._rand!()` situation, it seems like froma copula to the other there is no clear standard, I should investigate that. 

Correct your tests, and then we can move on to other models :)

Santymax98:
I manually calculated the values ​​for each of them and tried to put in a
fixed number of decimal values.

I took as reference each of the functions present in the texts. However,
could you take some examples or exercises to strengthen the tests a little?

When you set Base.rand() and Distributions.rand!() does it give any
ambiguity conflict?

I am going to implement the suggestion of the tests, it is the first time I
use them so I don't have much practice in it. After them, if you are
interested, you could continue implementing some Archimedean copulas and
try to include some extreme value copulas.

lrnv:
Well, This is probably the easiest thing to do yes, to create a new PR. You might add a few of them at once if you want. This PR interface allows us to discuss and collaborate easily on the code, dont you think ? 

I'll merge and close this one now, thanks a lot ! 

Btw, Congratulation on you first merged pull request !

PR #43, opened 2023-10-27
Plackett copula documentation

Santymax98:
Hi!!, i added information about the Plackett copula, let me know if anything else is missing


lrnv (commented):
The triple ticks documentation needs to be right before the object (i think). Otherwise this is perfect, thank you very much !

PR #46, opened 2023-10-30
Williamson transform sampling

lrnv:
Fot the moment this implementation is barebone... 
We also need documentation and bindings for a few known generators.  and a lot of tests. 

but otherwise this should work correctly. 

In the future, some copulas may leverage this implmentation, in partiuclar the version that provide both the radial dist and the generator (if both are known, better give both). 

We would also need the inverswe generator for some functions... maybe it can be provided through another field in the struct ? And/or computed automaticallly..


lrnv:
Exemples from Mcneil+Neshleova where the generator AND the williamson distribution are known analytically could be implemented as exemples

lrnv (commented):
Dont we need one more exponential to sample from de d-variate simplex ? I though so.

lrnv:
Closed in favor of #51

PR #51, opened 2023-11-06
Wiliamson transform sampling v2

lrnv:
Fot the moment this implementation is barebone...
We also need documentation and bindings for a few known generators. and a lot of tests.

but otherwise this should work correctly.

In the future, some copulas may leverage this implmentation, in partiuclar the version that provide both the radial dist and the generator (if both are known, better give both).

We would also need the inverswe generator for some functions... maybe it can be provided through another field in the struct ? And/or computed automaticallly..

Exemples from Mcneil+Neshleova where the generator AND the williamson distribution are known analytically could be implemented as exemples

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,7 +27,7 @@ Roots = "1, 2"
 SpecialFunctions = "2"
 StatsBase = "0.33, 0.34"
 TaylorSeries = "0.12, 0.13, 0.14, 0.15"
-WilliamsonTransforms = "0.0"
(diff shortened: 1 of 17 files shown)


lrnv:
@Santymax98 I think I have a propper implementation now using the williamson transform by default as we discuss by mail.

I will merge that tomorow (it is midnight in europe right now), if you want to take a look at it before ? Then starting tomorow once it is merged and published in the general registry we can start working on it to include your other archimedean models in the package :)

PR #53, opened 2023-11-08
Archimedean news

Santymax98:
With the purpose of extending the catalog of Archimedean copulas, I added two new copulas widely used in the literature for risk analysis, among others. I hope you find it useful.

### DIFF
--- a/docs/src/archimedean/available_models.md
+++ b/docs/src/archimedean/available_models.md
@@ -44,7 +44,7 @@ JoeCopula
 ## Gumbel-Barnet
 
 ```@docs
-GumbelBarnetCopula
+GumbelBarnettCopula
 ```
 
 
--- a/docs/src/archimedean/available_models.md
+++ b/docs/src/archimedean/available_models.md
@@ -39,3 +39,17 @@ AMHCopula
 ```@docs
 JoeCopula
 ```
+
+
+## Gumbel-Barnet
+
+```@docs
+GumbelBarnetCopula
+```
+
+
+## Inverve Gaussian
+
+```@docs
+InvGaussianCopula
+```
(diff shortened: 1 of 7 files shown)


lrnv (commented):
So : 

- I fixed a little the interface to comply with what the package uses, but there was really not much to do. 
- Is there some special cases from these copulas ? Some values of parameters that make them return the independence or one of the frechet-hoeffding bound ? If yes, then it would be nice to implement these in the generator directly (and comment about it in the docstring), see e.g. how it works for the clayton. 
- Could you add a few tests cases ? You can copy / leverage those from the other archimedeans for exemple. Typically, I'd like to be sure that sampling them, evaluating the cdf, the pdf, and the kendall tau are working (in the sense : not producing errors) AND right (in the sense: producing values that you derived by hand or that are derived somewhere else in a reference)

Once a few tests are added, this can be merged. Very good work !!

Santymax98:
@lrnv  Of course, I'm going to try to improve that.

lrnv:
@Santymax98 Thank you :) I also added them to the docs so that the CI does not fail. I love this WilliamsonTransforms, it makes adding "wild" archimedeans like your Gumbel-Barnett really easy :D

Santymax98:
I think Gumbel Barnett is an excellent example to see the power of
WilliamsonTransform

Santymax98:
I don't know why it gave an error, I think it was my fault for not knowing
how to use github well, however I tried to improve the writing a little to
standardize with all the others, another thing was that I corrected the
idependence of Frank's copula, it is independent when the parameter is zero
and before it was 1.

Santymax98:
As for relations to special cases of the inverse Gaussian copula, I will
continue studying them.

PR #57, opened 2023-11-11
Farlie - Gumbel - Morgenstern family Copula (FGM)

Santymax98:
modified:   src/ArchimedeanCopulas/InvGaussianCopula.jl
	modified:   src/Copulas.jl
	new file:   src/MiscellaneousCopulas/FGMCopula.jl
	modified:   src/MiscellaneousCopulas/PlackettCopula.jl
	new file:   test/fgmCopula_test.jl

Hello, I implemented this family of copulas, you may be interested in adding to the package, the copula has a lot of applicability and you can find a lot of information about it. Let me know what you think.


lrnv (commented):
Thanks ! A few remarks: 

- I removed the dependency on LinearAlgebra, only for `dot` this was not really needed. I also modified a few things, , including removing functions that were called only once, **so be sure to pull my modifications before re-working on it**.
- Be carrefull with type stability, avoid typing stuff as `Float64` but always use `T`. ( I did it for you).
- Avoid printing debug stuff in final code (I remove a print statement for you). 

Stuff to do before merging :

- Comments and variable name in non-english are not acceptable, please translate your code :)

- Creating vectors without giving their sizes by `Vector{T}()` and then `push!`ing in them is reaaaaly slow in julia, please remove these by computing the size beforehand and then indexing to insert values that will be much faster. 

- maybe a few combinatorics simplifications are possible ?

Santymax98:
Hello, I was reading your recommendations and the code definitely looks more compact and better presented, you are right with ``dot``. On the other hand, I had planned to calculate the size of the vector in the following way
size = sum(binomial(d, k) for k in 2:d)
coefficients_vector = Vector{Int}(undef, size)

and continue with the normal calculations. Finally, I tried to use a little theory to minimize the combinations, however it has been shown that in reality all possible combinations are needed to be able to find all the inequalities. The specific condition to verify the vectors is ``2^d`` inequalities which means that we need all those combinations.

You can see it in section 2 here,

https://link.springer.com/article/10.1007/s42081-021-00118-y

the section is very short. Let me know if you think that alternative would be useful to give size to the vectors.

On the other hand, I wanted to ask you about the ``check_dim()`` function, is it implemented somewhere in the package?

lrnv:
No I just implemented the check_dim function here for you, but since it looked general and I plan to eventually use it somewhere else in the package I moved it to the `Copula.jl` file. 

I understand the need for combinatorics, this is fine you are right. BTW, this paper proposes an estimation method, it could be nice to have too if it is not too hard to implement, under the `Distributions.fit()` method.

Santymax98:
Hello, I saw it and I couldn't study it, however I think that,
unfortunately, I will only be able to continue at work after Wednesday. ☹️

Still, I will try to advance some things

lrnv:
Take the time you need, no need to rush yourself ! We will merge this when you are ready :)

lrnv (commented):
Hey, 

I made a few changes, notably i simplified your code and reduced the alocation problems of the growing vectors. 

However, your tests are still passing but maybe i did break something that your tests did not catch ? 

I also fixed this check_dim issue by making it globally in all the package, now copulas have to define `_cdf` and not `cdf` anyumore. Distributions.jl already does that for `_logpdf` but apparently not for cdf of multivariate distributions, so I added it. 

One more thing for you to check out : is the definition of \rho right ? I loks liike it will return a vector ? should'nt it be a matrix and/or a single value ?

Santymax98:
Hello, sorry for the delay, actually the value of Rho is correct, because you only find it in the literature for the first parameter, anyone who is going to work with FGM should be clear about that limitation. It would not be a limitation of the package but of the theory of copulation itself. On the other hand, I wanted to tell you that the stochastic representation algorithm is much faster than the others. Currently I believe that Copulas.jl is the only package that has the implementation for all the parameters. In ``R``, to obtain samples with different parameters it uses an algorithm in C, the implementation of which I do not know.

Finally, I think this Copula is good visibility for the package because lately I am noticing quite a bit of movement in the literature regarding this copula and its extensions.

lrnv:
This branch starts to be huge, let's merge it now. Thanks @Santymax98  !

PR #67, opened 2023-11-16
Fix InvGaussianCopula when theta==0

### LINKED ISSUE #66: Inverse Gaussian Copula
Hi!!, 

In the Gaussian inverse copula when the parameter is zero it should return the independent copula (special case) as the verification is below the requirement that the parameter must be positive it is removing the opportunity to return the independent copula when the parameter is equal to zero

lrnv:
closes #66


lrnv:
let's merge that we'll fix compat entries in the next PR

PR #73, opened 2023-11-17
Archimedeans

Santymax98:
Hello, I think I fixed the equations and the calculation of the tau corresponding to these copulas.


lrnv:
Looks like the gumbelBarnett tau does not return anything ? the `result` line is commented

lrnv:
Thanks. Could you add a few coherence tests on these two functions ? e.g. test that tau(tauinv(x)) == x for two or three well-chosen x values.

Santymax98:
I'm going to work with that

Santymax98:
Hello @lrnv , I was trying to use the tests, however I am having a problem on my machine with the `quadgk` function telling me that it is not defined.

lrnv:
Yes you need to add `QuadGK` to the dependencies of the package ! While in the package repositoryin julia, use ``] add QuadGK` and it will modify the `Project.toml` file, that you can then commit. I will do it for you if you want but RN i'm on the bus back home, maybe tomorow

lrnv (commented):
I pushed to your branch with the fix for the missing package. But I still have weird behaviors in the tests... Cheking by hand, 

```
julia> Copulas.τ(GumbelBarnettCopula(2,0.7))
-0.2726863504658321
```

whihc is really weird since it is supposed to be only positive dependence for this copula ? 

-> The integrated function might be wrong ? 

Same thing for the InvGaussian.

Santymax98:
This copula represents negative dependence.
can you see, joe 2014 page 435

lrnv (commented):
So I marked the test as Broken... this makes it pass, but this is not good, something must be done about those tests to pass them into "non-broken" state again, this is not acceptable.

lrnv:
Indeed gumbelbarnett allows negative dependency, i corrected that sorry. But the tests are still broken :)

Santymax98:
I understand, what can we do? The copula is to model negative dependence,
unlike the Clayton copula, where negative dependence is not useful for
statistical modeling (since its density is zero in those cases). The Gumbel
Barnett copula is very important to deal with that type of dependence.

lrnv:
Now the tests are passing, I will merge this if @Santymax98  you agree ?

Santymax98:
Yes, I agree. Thank you for your help

lrnv:
My pleasure ! We fixed a lot of stuff with these tests that is really good.

PR #75, opened 2023-11-17
Moving generators to their own structs

lrnv:
This is not finished yet


lrnv:
Closed in favor of #79

PR #78, opened 2023-11-20
[Plotting] Add plots with plotly

Santymax98:
Hello, I decided to share some things I work with to visualize data and work with Copulas interactively. They are some graphs, let me know if you are interested in working with them and being able to do the same with `Plots.jl` (here they would be static). By the way, if you think they are useful, you could increase the information for their use and customization options in the 'readme'.

### DIFF
--- a/joss/paper.md
+++ b/joss/paper.md
@@ -8,16 +8,14 @@ authors:
     orcid: 0000-0002-7508-999X
     corresponding: true
     affiliation: 1
-  - name: YOUR NAME
-    orcid: 0000-0000-0000-0000
-    affiliation: "2, 3" # must use "" if you have more than one.
+  - name: Santiago Jimenez
+    orcid: 0000-0002-8198-3656
+    affiliation: 2 # must use "" if you have more than one.
 affiliations:
  - name: SESSTIM, Aix Marseille University, Marseille, France
    index: 1
- - name: YOUR FIRST AFFILIATION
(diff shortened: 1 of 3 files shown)


Santymax98:
I don't know why the `paper.md` was sent again.

lrnv:
End of day for me -- I'll take a look tomorrow. 

I would prefer Plots.jl if the change is not too hard to make.

I think a great solution would be to make this a package extension, that loads only if Plots.jl is loaded alongside Copulas.jl, that would be a very good interface.

Santymax98:
No problem, good rest, the codes are interactive because the graphs can
move and that is what I normally work with, however I don't think it will
be very difficult to change it to `Plots`

lrnv:
@Santymax98 I finally implemented plotting functionality, that you can see there using Plots.jl https://lrnv.github.io/Copulas.jl/dev/visualizations/ 

If this is enough for your purposes, you can close this PR. Also tell me if some other plot is still missing, or more generally what you think of the new plots.

Santymax98:
It looks good to me. I think the color contour plot, along with the scatter plot, shows how useful it is overall. I just think that, for documentation purposes, the Sklar surface plots should be more detailed. Notice that the numbers overlap on the axes. Perhaps the rotation angle could be changed slightly to better position them?

lrnv:
let me make an issue out of your comment, and htn close this one. Thanks for taking a look!

PR #84, opened 2023-11-22
Add PowerGenerator

lrnv:
This new generator shandles outer/inner power families of archimedean generators. 

A few tests and inclusion in the general documentaiton is still missing.


lrnv:
test faillure is unrelated, has already been fixed on master. Needs tests

lrnv (commented):
Lacks tests

PR #91, opened 2023-11-30
Raftery copula

Santymax98:
I added the multivariate Raftery copula and some tests. I did not implement Kendall's tau because it is a very long equation (its form is closed) however it is very long. If you think it is necessary, you could implement it. I leave you the main reference
[https://www.mdpi.com/2227-7390/11/2/414]

but you can also see details in `Nelsen 2006`.


src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,105 @@
...
+```
+
+
+where ``u_{(1)}, \\ldots , u_{(d)}`` denote the order statistics of ``u_1, \\ldots ,u_d``.
+
+More details about Multivariate Raftery Copula are found in :
+
+    Saali, T., M. Mesfioui, and A. Shabri, 2023: Multivariate Extension of Raftery Copula. Mathematics, 11, 414, https://doi.org/10.3390/math11020414.    
+    
+    Nelsen, Roger B. An introduction to copulas. Springer, 2006. Exercise 3.6.
```
lrnv, line 24: this line should be : 
```
[nelsen2006](@cite) Nelsen, Roger B. An introduction to copulas. Springer, 2006. Exercise 3.6. 
```

src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,105 @@
...
+            return IndependentCopula(d)
+        elseif θ == 1
+            return MCopula(d)
+        else
+            return new{d,typeof(θ)}(θ)
+        end
+    end
+end
+function _cdf(R::RafteryCopula, u::Vector{T}) where {T}
+    if length(u) != R.d
```
lrnv, line 45: R.d does not exists here. You may access the dimension of the copula using : 

```
function _cdf(R::RafteryCopula{d,P}, u::Vector{T}) where {d,P,T}
```

or by using `d = length(R)`, but `R.d` will not work. 

Moreover, there is no need to check the dimension of u this is done automatically now :)

src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,105 @@
...
+        throw(ArgumentError("Dimension mismatch"))
+    end
+    
+    # Order the vector u
+    u_ordered = sort(u)
+    
+    term1 = u_ordered[1]
+    term2 = (1 - R.θ) * (1 - R.d) / (1 - R.θ - R.d) * prod(u).^(1/(1 - R.θ))
```
lrnv, line 53: R.d does not exist. check everywhere i will not mark it everytime

src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,105 @@
...
+    u = rand(rng, dim+1)
+    
+    # Step 2: Generate j from a Bernoulli distribution with parameter θ
+    j = rand(Bernoulli(R.θ),1)
+    uj = u[1].^j
+    # Step 3: Calculate v_1, ..., v_d
+    for i in 2:dim+1
+        x[i] = u[i]^(1 - R.θ) * uj
+    end
+
```
lrnv, line 97: `x[1]` was not set ? and `x[d+1]` was set, while x only has d values so this will error.

src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,105 @@
...
+    end
+    
+    return x
+end
+
+function ρ(R::RafteryCopula{d,R}) where {d, P}
+    term1 = (d+1)*(2^d-(2-R.θ)^d)-(2^d*R.θ*d)
+    term2 = (2-R.θ)^d*(2^d-d-1) 
+    return term1/term2
+end
```
lrnv, line 105: So Kendall tau would be good indeed, even if the formula is long, but  most importantly is the maximum likelyhood procedure that is developped in the package : that would be nice to be able to fit this copula to some data !

test/RafteryTest.jl
```diff
@@ -0,0 +1,42 @@
...
+end
+
+@testitem "RafteryCopula Sampling" begin
+    using StableRNGs
+    rng = StableRNG(123)
+    n_samples = 100
+    F = RafteryCopula(3,0.5)
+    samples = rand(rng,F, n_samples)
+    @test size(samples) == (3, n_samples)
+end
```
lrnv, line 42: Could you add you copula, for one or two parameter value, into the list in `test/margins_uniformity.jl` ? this will automatically check that the samples are uniform and that the cdf is OK with these samples. This is not the best test, but it checks for coherence and usually catches a few bugs.

lrnv (approved):
Thanks a lot for this one, I did not even know this copula ! 

Overall this looks good, I gave a few comments on the code, if you could go fix these that would be very nice. Also a test of the correctness of the pdf and cdf would be nice.

There is something that tickles me : the current state of your code, especially the `R.d` error, is showing that you did not run your own tests before sending this code. How are you working with this package ? I suspect that you are complicating your own life by not using the standard workflow, maybe we could discuss that together if you want ?

Santymax98:
I work in vs code. Since I still don't have a good handle on Julia
packages, sometimes I need to make separate code and then join it together.


Abd… sometimes I delete the package on my computer and eat again to avoid.

I apologize if I'm making some mistakes.

Santymax98:
Thanks for the tip!

I'm going to learn because I want to try to make my own package.


On the other hand, don't forget to tell me if you think it's better to put
Kendall's tau or not.

Santymax98:
Perfect, I'll work on it.

lrnv:
So I merged the main branch into yours, so that there is no more conflicts. Dont forget to pull the modifications or you will erase them ! 

For the fitting procedure, if it is just a non-linear maximization of the LLH, then let's not implement it. We should implement that somewhere else as a generic, I agree.

src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,111 @@
...
+    j = rand(Distributions.Bernoulli(R.θ), 1)
+    uj = u[1]^j[1]
+    # Step 3: Calculate v_1, ..., v_d
+    for i in 2:dim+1
+        x[i-1] = u[i]^(1 - R.θ) * uj
+    end
+    
+    return x
+end
+function ρ(R::RafteryCopula{d,R}) where {d, P}
```
lrnv, line 93: There is an issue here reported by the tests: should be {d, P} and not {d,R} !

lrnv (changes requested):
Can you see the docs/src/assets/references.bib now ? 

Still a few tests but nothing too big i think, this is really good !

src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,111 @@
...
+    [Raftery2023](@cite) Saali, T., M. Mesfioui, and A. Shabri, 2023: Multivariate Extension of Raftery Copula. Mathematics, 11, 414, https://doi.org/10.3390/math11020414. 
+    
+    [nelsen2006](@cite) Nelsen, Roger B. An introduction to copulas. Springer, 2006. Exercise 3.6. 
+
+It has a few special cases:
+- When θ = 0, it is the IndependentCopula.
+- When θ = 1, it is the the Fréchet upper bound
+"""
+struct RafteryCopula{d, P} <: Copula{d}
+    θ::P  # Copula parameter
```
lrnv, line 31: Looks like the tests you implemented expects an argument error if you pass the dimension d=1, but your code does not throw in this case. Either remove the tests or include a check here that d >= 2.

lrnv (approved):

src/MiscellaneousCopulas/RafteryCopula.jl
```diff
@@ -0,0 +1,111 @@
...
+end
+function Distributions._logpdf(R::RafteryCopula{d,P}, u::Vector{T}) where {d,P,T}
+    # Order the vector u
+    u_ordered = sort(u)
+    
+    term_denominator = (1 - R.θ)^(d - 1) * (1 - R.θ - d)
+    term_numerator = 1 - d - R.θ * u_ordered[d]^((1 - R.θ - d) / (1 - R.θ))
+    term_product = prod(u)^((R.θ) / (1 - R.θ))
+    
+    logpdf_value = log(term_numerator) - log(term_denominator) + log(term_product)
```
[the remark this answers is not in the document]
Santymax98: I'm going to do it now

lrnv (approved):

lrnv:
Perfect ! Let's merge that then, congratulation ! I will tag a new release to push it into the registry

Santymax98:
Great! Thank you for the suggestions

PR #99, opened 2023-12-04
Update margins_uniformity.jl

lrnv:
to fix randomness


lrnv (commented):
another test without the raftery copula

PR #113, opened 2024-02-01
Fix typo -- IndependentCopula

pdeffebach:


lrnv:
Thanks a lot ! The docs indeed need a bit of proofreading...

Issue #120, opened 2024-02-04
[Paper] Typos to fix + Eventually automatic proof reading ? 


lrnv:
Link to JOSS review : https://github.com/openjournals/joss-reviews/issues/6189

I spotted a few typos, you may consider using a tool like  to check more systematically. This could also be setup on CI (actually the julia repo does it) but it might be an overkill.

- line 40 of the paper must -> most
- line 41 of the paper complience -> compliance
- line 46 of the paper specify -> specifies
- line 55 folloiwng -> following (I also think consists of is the grammatically correct form, instead of is consisting of but I am not native)
- line 63 exploit -> exploits
- Documentation navbar Exemples -> Examples
- [here](https://lrnv.github.io/Copulas.jl/stable/dependence_measures/) in the "Unfinished work" box, wtill -> still

lrnv:
See #123 for eventual automatic proof reading.

lrnv:
So completed now

PR #124, opened 2024-02-04
Fixes a few typos

lrnv:
This does not fix all of them, but it is a good way forward. 

Maybe we could implement some automatic checking.


lrnv (commented):
fixes #120

PR #171, opened 2024-02-23
Add examples

Santymax98:
Hello Oskar, I added an example with some important reference, I don't know what you think about it, if you need more just tell me. I closed because I forgot to create a new branch, is there a problem with that?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -31,7 +31,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: julia-actions/cache@v1
+      - uses: julia-actions/cache@v2
       - uses: julia-actions/julia-buildpkg@v1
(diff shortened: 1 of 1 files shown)


lrnv:
No problem wtih the fact that you do not have a feature branch. However, I think this should be a full example page on the documentation and not inside the paper, could you move it ? 

Look at this pull request here : https://github.com/lrnv/Copulas.jl/pull/169/files to see where to put your code so that it displays in the documentation. Beware, the syntax (in particular citing) is not exaclty the same, and the reference must be moved to the documentation references, not the paper one. 

Tip: While you are working locally on the documentation, you can view locally your chnages by: 

- launching a julia console 
- typing `] activate docs` to activate the docs in the pkg manager
- Then `using LiveServer` and then `servedocs()`. This will open a local web server on your machine so that you can check out the documentaiton you are writting

Santymax98:
I understand what you are asking of me, I saw the request in an email
earlier, I will move it

PR #174, opened 2024-02-23
Small typo

osorensen:
Since Archimedes was a person (although long ago), I think it's correct to use upper-case A throughout.

https://github.com/openjournals/joss-reviews/issues/6189


lrnv:
Indeed, Archimedean copula take an upper caps. Thanks :)

PR #211, opened 2024-08-03
Extreme value

Santymax98:
I try to repair documentation of extreme values


lrnv:
This brnach is full of conflicts ^^ 

I tried to fix it there #212 but we'll see in a few minuts if it worked or not. The fundamental issue is that there is no CI that compile the docs on PR's, this is something we could fix

lrnv:
Fixed in #214 :)

PR #219, opened 2024-09-18
Extreme value

Santymax98:
Hello Oskar, in this PR I try to solve something regarding the sampling of ``GalambosCopula`` you can notice that currently... if you take a parameter greater than 20... exactly 19.7 or greater... the function to obtain samples gives problems ...you can try this example
```julia
G = GalambosCopula(19.7)
rand(G, 1000)
```
You may notice that he gives an error. However, I saw that in the ``copula`` package of ``R`` the largest parameter that supports the generation of samples without error is approximately 72. In this sense I decided to implement an alternative. I added a warning to the structure and now it is possible to use larger parameter values... for example
```julia
G = GalambosCopula(210.0)
rand(G, 1000)
```


lrnv:
Hum... I guess that this needs a good rebase before being able to be merged. Let me try to do that

Santymax98:
Yes i can do it, do you think it is better to do it in another branch?

lrnv:
Yes, in another branch on top of an up-to-date main branch, that would be the easiest to handle

PR #220, opened 2024-09-19
Fix galambox

lrnv:
Should supplant #219 if @Santymax98 you agree ?

### DIFF
--- a/src/ExtremeValueCopulas/GalambosCopula.jl
+++ b/src/ExtremeValueCopulas/GalambosCopula.jl
@@ -26,12 +26,15 @@ References:
 struct GalambosCopula{P} <: ExtremeValueCopula{P}
     θ::P  # Copula parameter
     function GalambosCopula(θ)
+        if θ > 19.5
+            @warn "The parameter θ = $(θ) is large, which may lead to numerical instability in any functions. Consider regularizing the input."
+        end
         if θ < 0
             throw(ArgumentError("Theta must be >= 0"))
         elseif θ == 0
-            return IndependentCopula(2)
+            return IndependentCopula()
         elseif θ == Inf
-            return MCopula(2)
+            return MCopula()
         else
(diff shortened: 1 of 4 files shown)


Santymax98:
I just checked, it's perfect... I had tried it quite a bit. I'm a little busy... I think this is good. We should look for which values ​​of the parameters give numerical problems, for example for galambos it was approximately "19.6"... it would be interesting to see if the same thing happens with other types of copulas and determine in which parameters the generation of samples begins to fail to repair it. So far I can't find it but it would be interesting.

lrnv:
Maybe you could add here a test on the problematic parametrisation for galambos ?

Santymax98:
Are you referring to examples where the parameter is greater than 19.6? With this solution that we implemented we are getting good samples... for large parameters in fact for parameters greater than "75" which I believe is the maximum limit in the `copula` package of `R`.

lrnv:
Yes exactly, maybe we should write that down in the tests ? So that if, someday, we modify the algorithm and the samples are not as good, we'll know about it imediately

Santymax98:
understood. I'm going to do it...

lrnv:
Sorry I did not remember that you added an example in #219 ! I just added a test correpsonding to your example, is that enough in your oppinon ?

Santymax98:
yes... I was doing this...
```julia
@testitem "small extreme sampling test" begin
    using Copulas, Distributions, HypothesisTests
    
    params = [20, 60, 70, 80, 120, 210]
    for θ in params 
        cop = GalambosCopula(θ)
        data = rand(cop, 10^4)
        
        # Check data size
        @test size(data) == (2, 100)
    end
    @test true
end
```

lrnv:
I simply refactored the test files because they were very verbose. I'll wait for the online CI to go green and then I'll merge this. TY Santiago :)

Santymax98:
ok, it's perfect for me

PR #254, opened 2025-07-15
Add the rosenblatt transform and its inverse

### LINKED ISSUE #151: [New feature] Rosenblatt (and inverse) transformations
AnderGray said in Joss's review: 

> Rossenblat (and inverse) transformation is often used in uncertainty analysis (particularly reliability analysis) to transform random variables into independent standard normal variables.

----------------

There is a bit of rosenblatt transformation implemented in `BivariateCopulas.jl`. I know this transformation is very useful, I am just not really sure how we can make it generic enough to work for all copulas (on one side) and specific enough for standard models to obtain fast computations. 

It is also used to do quasi-random-number generation with a dependence structure, which is really useful for some use-cases.

FriesischScott:
This PR adds the (inverse) rosenblatt transform. I've implemented it for the `GaussianCopula` and all `ArchimedeanCopula` types.

The inverse rosenblatt for all archimedean copulas except the *Clayton* copula involves numerical root finding in all dimensions but the first. This is also what they do in the R *copula* package. Unfortunately no closed form solutions are available.

To define the transform I have updated and extended most of the generator definitions. For the *AMH*, *Clayton*, *Gumbel*, *Frank*, and *Joe* generatores I have added the required higher order derivatives of the generator provided in [1]. Where missing I have also added the appropriate types to the functions.

Since the *Clayton* copula has a closed form solution for the inverse rosenblatt I am using it to sample from the copula.

I have added extensive tests to ensure everything works as expected. Some of the exiting tests I had to remove. Since the original tests were checking for fixed random samples from all copulas in order, changing the `rand` function of the `ClaytonCopula` broke all following tests. Instead I am now testing that the resulting samples have the expected values of Kendall's tau and that the pdf integrates to 1.0.

The only real edge case is the `ClaytonCopula` with negative dependence. For the `cdf` this is easily handled by the existing `max(1+ t, 0)` generator definition. For the `pdf` it's slightliy more difficult because it involves the inverse generator and its derivative. I've opted to add a specific `_logpdf` method that excluded the region where the `pdf` is zero.  This was the only way I could properly get it to integrate to 1.0. It's a bit slower because of this but unless you compute the integral the speed should be sufficient.

The analytical derivatives required two additional dependencies. `BigCombinatorics` and `PolyLog`.  The existing uses of `Combinatorics` could potentially be replaced by `BigCombinatorics` as well to completely replace the dependency.

So far I have not written any documentation for the rosenblatt transform but if you want me to I'd be happy to add some docstrings and  section in the docs.

Finally, because you specify following `blue` style in the README I have added the matching `JuliaFormatter.toml`.  Resulting in quite a few unrelated formatting changes.

Closes #151 

---

 [1] M. Hofert, M. Mächler, and A. J. McNeil, ‘Likelihood inference for Archimedean copulas in high dimensions under known margins’, Journal of Multivariate Analysis, vol. 110, pp. 133–150, Sep. 2012, doi: [10.1016/j.jmva.2012.02.019](https://doi.org/10.1016/j.jmva.2012.02.019).


FriesischScott:
> So after a first glance at what you did it looks correct to me, thanks a lot for the efforts! Unfortunately, I would largely prefer to separate the formatting nonsense from the content itself in two PRs. Can you simply rebase your work to remove the formatting steps and only keep the content ? Will be easy if the formatting is in certain particular commits. 
> 
> Sorry about that :/
> 

No problem, I'll try to revert the purely formatting changes where possible.

lrnv:
Okay this is much more readable now :) Basically very great work, just one question : I do not understand what happend with the negative dependent clayton. Was'nt the williamson transform that we had before a better option than what you did here ? 

This will need documentation of course, and a bit more deep proof-reading by myself, but apart from that question this is good to go. GG.

lrnv:
By the way, maybe more on the theoretical level, do you think it would be possible to implement a *generic* version of the rosenblatt and inverse rosenblatt transformations ? A version taht, even if very inneficient, would work with any copula from the package ? If this is doable, i'd like to have that as well.

src/Generator/UnivariateGenerator/ClaytonGenerator.jl
```diff
@@ -9,13 +9,13 @@ Constructor
     ClaytonGenerator(θ)
     ClaytonCopula(d,θ)
 
-The [Clayton](https://en.wikipedia.org/wiki/Copula_(probability_theory)#Most_important_Archimedean_copulas) copula in dimension ``d`` is parameterized by ``\\theta \\in [-1/(d-1),\\infty)``. It is an Archimedean copula with generator : 
+The [Clayton](https://en.wikipedia.org/wiki/Copula_(probability_theory)#Most_important_Archimedean_copulas) copula in dimension ``d`` is parameterized by ``\\theta \\in [-1/(d-1),\\infty)``. It is an Archimedean copula with generator :
 
 ```math
 \\phi(t) = \\left(1+\\mathrm{sign}(\\theta)*t\\right)^{-1\\frac{1}{\\theta}}
```
lrnv, line 15: You should reflect the change of generator for the clayton in the docs

src/Generator/UnivariateGenerator/GumbelGenerator.jl
```diff
@@ -1,62 +1,68 @@
...
+    # first derivative of ϕ
+    a = 1/G.θ
+    tam1 = t^(a-1)
+    return - a * tam1 * exp(-tam1*t)
+end
+function ϕ⁽ᵏ⁾(G::GumbelGenerator, ::Val{d}, t) where d
+    α = 1 / G.θ
+    return ϕ(G, t) * t^(-d) * sum([
+        α^j * Stirling1(d, j) * sum([Stirling2(j, k) * (-t^α)^k for k in 1:j]) for j in 1:d
+    ])
```
lrnv, line 52: You dont need to instantiate the vector to sum it, simply use `sum(... for i in I)` instead of `sum([... for i in I])`, its better for perfromances

lrnv: Actually you did that a few times across the PR so please check them

src/ArchimedeanCopula.jl
```diff
@@ -48,45 +48,34 @@ struct ArchimedeanCopula{d,TG} <: Copula{d}
...
-    for us in u
-        ϕ⁻¹u = ϕ⁻¹(C,us)
-        sum_ϕ⁻¹u += ϕ⁻¹u
-        logdenom += log(-ϕ⁽¹⁾(C,ϕ⁻¹u)) # log of negative here because ϕ⁽¹⁾ is necessarily negative
-    end
-    numer = abs(ϕ⁽ᵏ⁾(C, sum_ϕ⁻¹u))
-    if numer > 0
-        r = log(numer) - logdenom
-        !isnan(r) && return r
+    return log(ϕ⁽ᵏ⁾(C, Val(d), sum(ϕ⁻¹.(C, u))) * prod(ϕ⁻¹⁽¹⁾.(C, u)))
```
lrnv, line 63: Is'nt that a bit wastefull to compute separately `ϕ⁻¹.(C, u)` and `ϕ⁻¹⁽¹⁾.(C, u)` ? I mean, when there are special methods for both, its certainly OK, but when we go through the default case (automatic derivation) then its wastefull: a call to `WilliamsonTransforms.taylor.(Ref(x -> ϕ⁻¹(C,x)), u , Val(1))` will return the same as `zip(ϕ⁻¹.(C, u), ϕ⁻¹⁽¹⁾.(C, u))`, but with half as much computations: maybe we could try to leverage that ?

src/ArchimedeanCopula.jl
```diff
@@ -48,45 +48,34 @@ struct ArchimedeanCopula{d,TG} <: Copula{d}
...
+end
+
+function Distributions._logpdf(C::ArchimedeanCopula{d,TG}, u) where {d,TG<:ClaytonGenerator}
+    if !all(0 .< u .< 1) || (C.G.θ < 0 && sum(u .^ -(C.G.θ)) < (d - 1))
+        return eltype(u)(-Inf)
     end
-    return -T(Inf)
+    return log(ϕ⁽ᵏ⁾(C, Val(d), sum(ϕ⁻¹.(C, u))) * prod(ϕ⁻¹⁽¹⁾.(C, u)))
```
lrnv, line 70: Same thing here, the two calls are probably computing a lot of similar stuffs.

src/ArchimedeanCopula.jl
```diff
@@ -198,3 +187,101 @@ function Distributions._rand!(rng::Distributions.AbstractRNG, ::ArchimedeanCopul
...
+
+function rosenblatt(C::ArchimedeanCopula{d,TG}, u::AbstractMatrix{<:Real}) where {d,TG}
+    @assert d == size(u, 1)
+
+    U = zeros(eltype(u), size(u))
+    U[1, :] = u[1, :]
+
+    for j in 2:d
+        U[j, :] .=
+            ϕ⁽ᵏ⁾.(C.G, Val(j - 1), reduce(+, [ϕ⁻¹.(C.G, u[k, :]) for k in 1:j])) ./ ϕ⁽ᵏ⁾.(C.G, Val(j - 1), reduce(+, [ϕ⁻¹.(C.G, u[k, :]) for k in 1:(j - 1)]))
```
lrnv, line 207: Here too, the syntax `reduce(+, [... for k in K])` would be more efficilty written as `sum(... for k in K)`, to avoid allocation.

src/ArchimedeanCopula.jl
```diff
@@ -198,3 +187,101 @@ function Distributions._rand!(rng::Distributions.AbstractRNG, ::ArchimedeanCopul
...
+    @assert d == size(u, 1)
+
+    U = zeros(eltype(u), size(u))
+    U[1, :] = u[1, :]
+
+    for j in 2:d
+        U[j, :] .=
+            (
+                (1 .- j .+ sum(u[1:j, :] .^ (-C.G.θ); dims=1)[:]) ./
+                (2 .- j .+ sum(u[1:(j - 1), :] .^ (-C.G.θ); dims=1)[:])
```
lrnv, line 225: Here you are recomputing `u^θ` a lot of times. maybe pre-compute it ?

src/ArchimedeanCopula.jl
```diff
@@ -198,3 +187,101 @@ function Distributions._rand!(rng::Distributions.AbstractRNG, ::ArchimedeanCopul
...
+    U = zeros(eltype(u), size(u))
+    U[1, :] = u[1, :]
+
+    for i in axes(u, 2)
+        for j in 2:d
+            f =
+                x ->
+                    (
+                        ϕ⁽ᵏ⁾(C.G, Val(j - 1), sum(ϕ⁻¹.(C.G, [U[1:(j - 1), i]..., x]))) /
+                        ϕ⁽ᵏ⁾(C.G, Val(j - 1), sum(ϕ⁻¹.(C.G, U[1:(j - 1), i])))
```
lrnv, line 251: Here too there is a vector `[U[1:(j - 1), i]..., x]``that could simply be replaced by a tuple `()` or even you could replace `sum(ϕ⁻¹.(C.G, [U[1:(j - 1), i]..., x])))` by `sum(ϕ⁻¹.(C.G, U[1:(j - 1), i]))+ϕ⁻¹(x)` for cleanliness

src/ArchimedeanCopula.jl
```diff
@@ -198,3 +187,101 @@ function Distributions._rand!(rng::Distributions.AbstractRNG, ::ArchimedeanCopul
...
+                (2 .- j .+ sum(u[1:(j - 1), :] .^ (-C.G.θ); dims=1)[:])
+            ) .^ (-1 / C.G.θ - (j - 1))
+    end
+    return U
+end
+
+function rosenblatt(C::ArchimedeanCopula{d,TG}, u::AbstractVector{<:Real}) where {d,TG}
+    @assert d == size(u, 1)
+
+    return rosenblatt(C, reshape(u, (d, 1)))[:]
```
lrnv, line 234: You should `reshape()` instead of `[:]` to avoid re-allocating the space

src/ArchimedeanCopula.jl
```diff
@@ -198,3 +187,101 @@ function Distributions._rand!(rng::Distributions.AbstractRNG, ::ArchimedeanCopul
...
+
+    U = zeros(eltype(u), size(u))
+    U[1, :] = u[1, :]
+
+    for j in 2:d
+        U[j, :] =
+            (
+                1 .+
+                (1 .- (j - 1) .+ sum(U[1:(j - 1), :] .^ -C.G.θ; dims=1))[:] .*
+                (u[j, :] .^ (-1 / (j - 1 + 1 / C.G.θ)) .- 1)
```
lrnv, line 274: Once again, i feel likeyou are computing these powers several times ? Maybe i'm wrong

src/Copulas.jl
```diff
@@ -16,6 +16,8 @@ module Copulas
     import LogExpFunctions
     import QuadGK
     import LinearAlgebra
+    using PolyLog
+    using BigCombinatorics
```
lrnv, line 20: Please use `import` only and qualify your uses of the functions when you do use them, this improves readability of the code

lrnv (commented):
Here are a few comments on your code, nothing that cannot be solved easily :)

src/ArchimedeanCopula.jl
lrnv: And also you are computing the whole matrix of `ϕ⁻¹.(C.G, u)` twice, you could compute it on another line only once

src/ArchimedeanCopula.jl
lrnv: Even without recomputing u^\theta, still some of these sum looks like too much work

src/ArchimedeanCopula.jl
lrnv: In fact yes, the `sum(ϕ⁻¹.(C.G,...)))` could be precomputed before instantiating the function to reduce its cost drastically.

lrnv:
No problem, i'm working a bit on performance issues now

FriesischScott:
> Also in BivariateCopulas.jl there are two extra tests on the rosenblatt transfrom (see there https://github.com/AnderGray/BivariateCopulas.jl/blob/1ca144048f76cf782f9fb7c6b843ad5decfc2133/test/copulas.jl#L51-L64 ):
> 
>     1. test that the kendall tau of rosenblatt(C, rand(C,n)) is approximately zero (should be independent.
> 
>     2. test that rosenblatt(C, inverse_rosenblatt(C, u)) == u, that is that the two operations are indeed inverse of each other.
> 
> 
> It would be nice to add these two tests for several different copulas to ensure a certain coverage and validte even more our implementation.

The test for the Clayton copulas got lost in the merge but I brought them back now. All Archimedean and the Gaussian copula should now have both these tests.

FriesischScott:
> @FriesischScott Sorry i was trying to repond to you but basically deleted your response...
> 
> So you are saying the willimaosn transform of the clayton copula is not the right one anymore ? Because you changed the generator ?

That could be. I'm not familiar enough with the williamson transform to confirm but when you comment out the special `rand` methods for the Clayton copula and run the tests you should be able to see where it goes wrong.

lrnv:
@FriesischScott So what do you think of the Clayton now ?

Apart from that looks ok we could move to documentation and then merge

lrnv (approved):
This for me is good to go.

FriesischScott:
I've had a look now and I would prefer to go with my implementation of the Clayton Copula for a couple of reasons.

I believe the pdf should be nonzero on the discontinuous border for negative dependence.  As you can see in the tests for `C = ClaytonCopula(2,-0.5)` yours is `0.0` and mine is `2.0`. This is debatable and depends on if it's left or right continuous but this is the least of the issues I see.

It uses the standard version of the generator that everyone is familiar with except for the added `max` to make the `cdf` work. It's probably best to go with the established definition but refer to the paper for the proof that d-monotonicity is a sufficient requirement for negative dependence. The code is also easier to read without the conditional statements in the generator functions.

Most importantly it's significantly faster probably because of not having to use numerical differentiation in the Williamson transform. I've run a simple benchmark like this

```julia
C = ClaytonCopula(2,-0.5)
@benchmark rand($C, 10^6)
```

and this is the output using the inverse rosenblatt sampling.

```
BenchmarkTools.Trial: 64 samples with 1 evaluation.
 Range (min … max):  56.380 ms … 161.966 ms  ┊ GC (min … max):  8.22% … 61.07%
 Time  (median):     61.309 ms               ┊ GC (median):     7.87%   
 Time  (mean ± σ):   78.885 ms ±  36.990 ms  ┊ GC (mean ± σ):  28.21% ± 21.68%

  ▄█▇▅
  ████▄▄▁▁▃▃▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▄▁▃▁▄▄▄▃ ▁        
  56.4 ms         Histogram: frequency by time          161 ms <

 Memory estimate: 106.81 MiB, allocs estimate: 33.
```

and this is the result from sampling using the Williamson transform

```
BenchmarkTools.Trial: 1 sample with 1 evaluation.
 Single result which took 10.779 s (10.41% GC) to evaluate,
 with a memory estimate of 7.53 GiB, over 271352446 allocations.
```

 What do you think?

lrnv (commented):
No, i guess its the root solving because we did not specialize the inverse rosenblatt transformation for the clayton. I guess that even from by generator the root solving can be solved by hand to obtain a fast function. Let me try that.

FriesischScott:
Wait it looks like it's dispatching to the general implementation not the specific Clayton function.

FriesischScott:
> The williamson transform was wrong with your generator :/

With sampling from the closed form solution of the inverse rosenblatt  what do weneed the williamson transform for?

lrnv (commented):
@FriesischScott Could you propose a modification that reinstore the closed form solution for the inverse rosenblatt, but without modifying the generator defintion and the rest (only the `inverse_rosenblatt(C::ClaytonCopula{d}, u::AbstractMatrix{<:Real}) where {d}` function) ? I have the intuition that this should be enough.

FriesischScott:
> @FriesischScott Could you propose a modification that reinstore the closed form solution for the inverse rosenblatt, but without modifying the generator defintion and the rest (only the `inverse_rosenblatt(C::ClaytonCopula{d}, u::AbstractMatrix{<:Real}) where {d}` function) ? I have the intuition that this should be enough.

Will do 👍

lrnv:
The performance of the samplers is something that should be reviewed all around the package... but maybe later. Let's merge this once the Ci online tests are OK.

PR #265, opened 2025-08-30
a little api fit

Santymax98:
Hello dear Oskar, I've been working on some things and thought this might be helpful... Let me know if you're okay with continuing with the documentation and removing some of the Spanish comments...


lrnv:
Santiago, 

This is a *huge* piece of work, over 4000 lines. First, thanks a lot. Second, we cannot merge this like that I am sorry: can you split this into several pull-requests ? One PR per new functionality, one more per new copula would be the right amount. This will allow us to discuss things one by one and not all at once. 

You can keep the big branch and keep rebasing on top of the new main until there is nothing more in it if you want to be certain that we got everything. or whatever other processus you propose. 

Let's discuss. By the way, great work :rofl:

Santymax98:
Of course, I thought the same thing. I'll break it down... I also thought it was a big job... I'll organize it into small parts. Are you okay with that? Maybe that way I can include more specific documentation and testing. 

Little by little we can put a good update do you agree with them?

lrnv:
I just put that here as it might be useful when we deal with the fitting interface: https://discourse.julialang.org/t/automatic-differentiation-of-loglikelihood-function-am-i-doing-it-right/132157/4?u=lrnv

lrnv:
So @Santymax98 What do we do, do we rebase this PR or do you do another one ? Since its the end right we are at the fittin gsystem now ,

PR #266, opened 2025-08-31
Add BB1, BB2, BB3, BB6, BB7, BB8, BB9 and BB10 Copulas

Santymax98:
Hi, for the first time, I decided to include these two-parameter Archimedean copulas. Some have multivariate extensions, but I think they require a bit more work. For now, I think this is the most useful.

### DIFF
--- a/src/EllipticalCopula.jl
+++ b/src/EllipticalCopula.jl
@@ -75,3 +75,7 @@ function make_cor!(Σ)
         end
     end
 end
+
+function Base.show(io::IO, C::EllipticalCopula)
+    print(io, "$(typeof(C))(Σ = $(C.Σ)))")
+end
--- a/src/MiscellaneousCopulas/EmpiricalCopula.jl
+++ b/src/MiscellaneousCopulas/EmpiricalCopula.jl
@@ -32,6 +32,9 @@ function EmpiricalCopula(u;pseudo_values=true)
     end
     return EmpiricalCopula{d,typeof(u)}(u)
 end
+function Base.show(io::IO, C::EmpiricalCopula)
+    print(io, "EmpiricalCopula{d}$(size(C.u))")
+end
 function _cdf(C::EmpiricalCopula{d,MT},u) where {d,MT}
    return sum(all(C.u .<= u,dims=1))/size(C.u,2) # might not be very efficient implementation. 
 end
--- a/src/MiscellaneousCopulas/FGMCopula.jl
+++ b/src/MiscellaneousCopulas/FGMCopula.jl
@@ -51,6 +51,9 @@ struct FGMCopula{d, Tθ, Tf} <: Copula{d}
         return new{d, typeof(vθ), typeof(fᵢ)}(vθ, fᵢ)
     end
 end
+function Base.show(io::IO, C::FGMCopula{d, Tθ, Tf}) where {d, Tθ, Tf}
+    print(io, "FGMCopula{$d}(θ = $(C.θ))")
+end
 Base.eltype(C::FGMCopula) = eltype(C.θ)
 function _fgm_red(θ, v)
     # This function implements the reduction over combinations of the fgm copula. 
--- a/src/MiscellaneousCopulas/SurvivalCopula.jl
+++ b/src/MiscellaneousCopulas/SurvivalCopula.jl
@@ -28,6 +28,9 @@ struct SurvivalCopula{d,CT,VI} <: Copula{d}
         return new{d,typeof(C),typeof(indices)}(C,indices)
     end
 end
+function Base.show(io::IO, C::SurvivalCopula)
+    print(io, "SurvivalCopula($(C.C))")
+end
 function reverse!(u,idx)
     if ndims(u) == 1
         for i in idx


Santymax98:
@lrnv I checked and the error is found by aqua.jl. Do you think you know what's going on? Thanks for your help.

lrnv:
@Santymax98  The aqua things are not a big issue i think, not our fault. On your code, I have a few remarks / did a few modifications: 

1) These generators are constructed from pure frailty models, and thus are suitable for any dimensions of Archimedean copulas => I removed the dimension cap. 

2) No need to define one-time-use functions such as ϕ⁽²⁾ or frailty_dist, i removed them. 
3) Functions that are specific to one model, such a \eta in BB8 should have an underscore in front of their name. 

4) Regarding the `Distributions.params`, I left it as it was for the moment, I suspect it wil be usefull when the fitting comes. 

5) Each univariate distributions used as frailties should be in its own file, i created the files. 

6) On the testing side: 
		- Remember that you can use the `runtests.jl` file locally to filter which tests you run. I added a `:BBCops` tag for that.
		- I added your examples to generic testing. 
		- I fixed a lot of bugs.
		- Maybe we could alos make generic / specific tests in dimensions d= 3 or d= 4 ? 

I have a few errors: 
1) The rosenblatt / inverse rosenblatt routines fails for BB1 & BB2 & BB3. This usually indicates that there is an issue in the generators and their derivatives, usually in the limits. 
2) Integration of the pdf for BB2 gives NaN, for BB3 gives nonsensical values, for BB7 gives Inf.  This might NOT be an issue if these copula are not absolutely continuous (the integrator is not very smart when there is a discrete component). If this is the case, we can just excluse them. Otherwise, we might want to check again the pdf definition for these. 

Please do : 
- Use  `filter=ti->(:BBCops in ti.tags)` in your runtests.jl to run the tests and try to fix them. 
- Tell me what you think about the errors that i have. 
- If you specific tests are duplicating tests from the generic ones, please remove them no nee to test things twice. 
- If your speciic tests check for a property that is more generic than only the familly, move it in the generic tests to also test other models. 
- At the end, run the generic tests on models of higher dimensions to see what is happening.


Waiting for your input :)

Santymax98:
Oskar, yes, it is natural, the multivariate extension can be used without any problem, something strange happens with the others. I have reviewed the BB7 component and it works well for me. As for BB3 and BB2, I will go into a little more detail about BB3 and BB2, the problem always occurs when the values ​​are close to 0... in this case the PDF is going to explode.

I will try to improve it 👍

Santymax98:
Yes, I also tested in higher dimensions and the tests work well except in BB1 the Rosenblatt functions but if you look closely at the values ​​you will see that only one or two values ​​of all the results fail... the rest have a pretty good approximation. 

As for specific tests, I generally tried to look for the same ones that were implemented before this update. I only added PDF verification via MC integration and a small property on the CDF.

Santymax98:
For BB2 with $\theta=3,\ \delta=2$ (bivariate), using

$$
\varphi(s)=(1+\tfrac{1}{\delta}\log(1+s))^{-1/\theta},\quad
\varphi^{-1}(t)=\exp\!\big(\delta(t^{-\theta}-1)\big)-1,
$$

the density is

$$
c(u,v)=\varphi''(s)\,(\varphi^{-1})'(u)\,(\varphi^{-1})'(v),\qquad s=\varphi^{-1}(u)+\varphi^{-1}(v),
$$

with

$$
\varphi''(s)=ab\,(1+s)^{-2}\,g(s)^{-b-2}\,\big(g(s)+(b+1)a\big),\quad
g(s)=1+a\log(1+s),\ a=\tfrac{1}{\delta},\ b=\tfrac{1}{\theta},
$$

and

$$
(\varphi^{-1})'(t)=e^{\delta(t^{-\theta}-1)}\;\delta\theta\;t^{-(\theta+1)}.
$$

Let's suppose $(u,v)=(0.01,\,0.4)$

* $y_1=\delta(0.01^{-3}-1)\approx 1{,}999{,}998$, $y_2=\delta(0.4^{-3}-1)=29.25$.
* $s\approx e^{y_1}$ (dominated by $u=0.01$), $\log(1+s)\approx y_1$,
$g(s)\approx 1+\tfrac{1}{2}y_1\approx 10^6$.

Working in the **log domain** (to avoid $0\cdot\infty$) we get:

$$
\log c(u,v)\approx -1{,}999{,}963.293\quad\Rightarrow\quad
c(u,v)\approx 9.50\times 10^{-868\,574}.
$$

```julia

**Comment by lrnv** [comment]
Waouh this is wild. If you have other values you obtained by hand, add the bigfloat computation as a test for the few examples you have, it'll help us find the right values. But the pdf should really give a 0 here not a NaN if it behaved corectly..

**Comment by Santymax98** [comment]
if we use

```julia
function logpdf_bb2(G::BB2Generator, u1::T, u2::T) where {T<:AbstractFloat}
    θ, δ = G.θ, G.δ
    a, b = inv(δ), inv(θ)

    # Xi = δ*(u^{-θ}-1) usando expm1 para estabilidad
    X1 = δ * expm1(-θ*log(u1))
    X2 = δ * expm1(-θ*log(u2))

    # L = log(1 + φ⁻¹(u1) + φ⁻¹(u2)) = log( exp(X1)+exp(X2) - 1 )   (sin overflow)
    m = max(X1, X2, zero(T))
    L = m + log(exp(X1-m) + exp(X2-m) - exp(-m))    # exp(-m)≈0 si m≫1

    # g(s) = 1 + a*L  ⇒  log g = log1p(a*L)
    logg = log1p(a*L)

    # log φ''(s) = log(a)+log(b) - 2L + (-b-1)log g + log1p((b+1)*a / g)
    logφpp = log(a) + log(b) - 2L + (-b-1)*logg + log1p((b+1)*a*exp(-logg))

    # log (φ⁻¹)'(u) = δ(u^{-θ}-1) + log(δθ) - (θ+1)log u
    logd1 = X1 + log(δ*θ) - (θ+1)*log(u1)
    logd2 = X2 + log(δ*θ) - (θ+1)*log(u2)

    return logφpp + logd1 + logd2
end

function _pdf(B::BB2Copula, u::AbstractVector{T}) where {T<:AbstractFloat}
    ε = max(sqrt(eps(T)), T(1e-12))
    u1 = clamp(u[1], ε, 1-ε)
    u2 = clamp(u[2], ε, 1-ε)
    ℓ = logpdf_bb2(B.G, u1, u2)
    return exp(ℓ)
end
```

We get

```julia
julia> B = BB2Copula(2, 3, 2.0)
BB2Copula{2, Int64}(3, 2)

julia> Copulas._pdf(B, [0.01, 0.4])
0.0
```

But I'm not sure because there is an error with the generic one.

Santymax98:
If we use
```julia
function _logpdf(G::BB2Generator, u1::T, u2::T) where {T<:AbstractFloat}
    θ, δ = G.θ, G.δ
    a, b = inv(δ), inv(θ)
    X1 = δ * expm1(-θ*log(u1))
    X2 = δ * expm1(-θ*log(u2))
    m = max(X1, X2, zero(T))
    L = m + log(exp(X1-m) + exp(X2-m) - exp(-m))
    logg = log1p(a*L)
    logφpp = log(a) + log(b) - 2L + (-b-1)*logg + log1p((b+1)*a*exp(-logg))
    logd1 = X1 + log(δ*θ) - (θ+1)*log(u1)
    logd2 = X2 + log(δ*θ) - (θ+1)*log(u2)

    return logφpp + logd1 + logd2
end

function _pdf(B::BB2Copula, u::AbstractVector{T}) where {T<:AbstractFloat}
    ε = max(sqrt(eps(T)), T(1e-12))
    u1 = clamp(u[1], ε, 1-ε)
    u2 = clamp(u[2], ε, 1-ε)
    ℓ = _logpdf(B.G, u1, u2)
    return exp(ℓ)
end
```

We get

```julia
julia> B = BB2Copula(2, 3, 2.0)
BB2Copula{2, Int64}(3, 2)

julia> Copulas._pdf(B, [0.01, 0.4])
0.0
```
But I'm not sure because there is an error with the generic Distributions.jl

Santymax98:
Hi @lrnv, I reviewed the PDFs and their integration and concluded that the error ultimately lies in the integration method. If we use a basic Monte Carlo integration, it fits well, for example.

```julia
julia> function mc_pdf_integral(C; N::Int=2*10^6, seed::Integer=42)
           rng = Random.seed!(seed)
           d = length(C)
           u = Vector{Float64}(undef, d)

           logS  = -Inf
           logS2 = -Inf
           n_eff = 0

           @inbounds for _ in 1:N
               rand!(rng, u)
               lp = logpdf(C, u)
               if isfinite(lp)
                   logS  = LogExpFunctions.logaddexp(logS,  lp)
                   logS2 = LogExpFunctions.logaddexp(logS2, 2lp)
                   n_eff += 1
               end
           end
           n_eff == 0 && error("mc_pdf_integral: There were no finite evaluations.")

           Ĩ  = exp(logS  - log(n_eff))
           m2 = exp(logS2 - log(n_eff))
           var̂ = max(m2 - Ĩ^2, 0.0)
           SE = sqrt(var̂ / n_eff)
           return Ĩ, SE, n_eff
       end
mc_pdf_integral (generic function with 1 method)
julia> B = BB2Copula(2,1.2,0.5)

BB2Copula{2, Float64}(1.2, 0.5)

julia> mc_pdf_integral(B)
(0.9942638445339282, 0.004312210810533557, 2000000)
```
Using `hcubature` increasing the maxevals a little... I got the following
```julia
julia> v, r = hcubature(x -> pdf(B, x), zeros(2), ones(2), maxevals=1000000)
(0.9843750001362106, 1.4665443295334496e-8)
```

For `BB3Copula` it is much more unstable but by adjusting some values ​​I managed to get good results. Do you think we can use the `mc_pdf_integrate` function only when the copula is BB2 or BB3 and let the others integrate with hcubature?

lrnv:
Yes that would be the best solution, use the cheap version when the copula un well behaved and the stronger version overwise. Maybe you can code It up like that : 

1. Perform the hcubature 
2. If the value is not the right one, retry with monte-carlo
3. If the value is still not right, then the test did not pass, report it. 

So that our approach is agnostic to the copula and works for everything

lrnv:
@santymax you reverted the changes I had done on generators of BB2 and BB3, which is why you have issues with rosenblatt (these changes solved them). 

Look at the diff of you last commit ( the merging one) and cancel everything that has to do with the generators in those files and you should be good to go

Santymax98:
Yes, I was doing it, but I still haven't passed a single test. In a single test

lrnv:
So, @Santymax98 I rewrote a bit your methods too simplify them, and solved the issue on the BBcops. On the other hand, you had commit the restriction of tests to themonly : I removed it, and now a looooot of other copulas do not pass the intergation tests ^^ In particular the second one. Now the question is what the fuck is happening with the integration test, why so many copulas dont pass through it...

I will continue to investigate, but if you have an idea tell me :)

lrnv:
So in fact, it was only happening in dimension 3... because the `measure` function we were testing about was wrong ><"

I fixed it and now it should work correctly on everything... Let's see what the online test turns out with

lrnv:
@Santymax98 So now all tests passes. Let me review a bit the code, and then we can merge, and I'll wait for you next PR :)

Santymax98:
@lrnv This PR was quite a bit of work, jajaja 🤣i’m sorry, but thanks for your support. I don't think any other package has this much work.

PR #286, opened 2025-09-12
[New Copulas] d > 2 EV and Archimax

lrnv:
I am not convinced about this begining of implementation...


lrnv:
Closing this exploratory draft: it conflicts with current main, its design was explicitly uncertain, and the EV/Archimax internals have evolved substantially since it was opened. A future d > 2 implementation should start from current main with a focused design and validation rather than rebase this prototype.

PR #295, opened 2025-09-16
Adding empirical copulas

Santymax98:
Hi Oskar, in the spirit of moving forward with the fit API, I decided to include these important empirical copulas known in the literature... I believe they're also in the Roadmap. However, I'm having some issues with the tests.

1. In `BernsteinCopula` and `BetaCopula`, I don't know which tests are broken.
2. In `EmpiricalEVCopula`, anything that implements the second derivative of A will fail, such as pdf and the Rosenblat functions. This is because it's a discrete function, and the second derivative will return zero.
3. In `CheckerboardCopula`, I noticed that some tests don't pass; however, I think it's a numerical problem. It should be 1, but we get `0.999999999`. In this case I tried to implement it following this representation https://www.degruyterbrill.com/document/doi/10.1515/demo-2020-0004/html

Do you think you could take a look?

### DIFF
--- a/docs/src/assets/references.bib
+++ b/docs/src/assets/references.bib
@@ -992,3 +992,44 @@
   year={2014},
   publisher={Elsevier}
 }
+
+@article{segers2017empirical,
+  title={The empirical beta copula},
+  author={Segers, Johan and Sibuya, Masaaki and Tsukahara, Hideatsu},
+  journal={Journal of Multivariate Analysis},
+  volume={155},
+  pages={35--51},
+  year={2017},
+  publisher={Elsevier}
+}
+
+@article{sancetta2004bernstein,
+  title={The Bernstein copula and its applications to modeling and approximations of multivariate distributions},
+  author={Sancetta, Alessio and Satchell, Stephen},
+  journal={Econometric theory},
+  volume={20},
+  number={3},
+  pages={535--562},
+  year={2004},
+  publisher={Cambridge University Press}
+}
+
+@article{gudendorf2011nonparametric,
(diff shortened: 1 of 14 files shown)


lrnv:
I will take a look a soon as I can, which is probably on Saturday sorry for the delay.

Do you know that I implemented checkerboard copulas in R a long time ago ? Among other models ;) See https://lrnv.github.io/cort

Santymax98:
Yes, in fact, that's where I got the idea from. That link has some very good information.

On the other hand, Bernstein's computational cost is high compared to the others. I tried precomputing some things but couldn't see what... I think the beta case isn't too difficult, and as for Checkerboard, I'll review it... In general, I tried to follow the idea of ​​the article, but I think your recommendations are much more efficient.

Santymax98:
Hi @lrnv, I was reviewing the checkerboard from data and noticed:

**Construction from data.** Given pseudo-observations $U^{(1)},\dots,U^{(n)}\in(0,1)^d$ and a regular grid with $m=(m_1,\dots,m_d)$, the empirical counts $\hat h_k=\\#\{U^{(r)}\in B_k\}/n$ satisfy $\sum_k \hat h_k=1$ but, in general, not $\sum_{k:\,k_i=t}\hat h_k=1/m_i$.
To obtain a valid **checkerboard copula** (uniform margins), we **should project** $\hat h$ onto the multi-stochastic set
$$\{\,h\ge0:\ \sum_k h_k=1,\ \ \sum_{k:\,k_i=t} h_k = 1/m_i\ \forall i,t\,\}$$
(e.g., via IPF/Sinkhorn). This preserves positivity and enforces the exact slice sums or not? I have this little doubt

Issue #301, opened 2025-09-18
Clayton Copula Theta inference error


rand5:
MWE for inferring theta when using the Gumbel copula:
```
@model function CoreFun(F;N)

    θ ~ Gamma(1.1,0.75)+1.0
    y ~ filldist(SklarDist(GumbelCopula(2,θ),(Uniform(), Uniform())), N)
    
    return y
end

CoreFun(y::Matrix{<:Real}) = CoreFun(F; N=size(y,2)) | (; y)

model = CoreFun(rand(2,1000))

sampler = NUTS();
chain = sample(model, sampler, 4_000, progress=false)
```
runs smoothly.

Switching to the Clayton copula: 
```
@model function CoreFun(F;N)

    θ ~ Gamma(1.1,0.75)-1.0
    y ~ filldist(SklarDist(ClaytonCopula(2,θ),(Uniform(), Uniform())), N)
    
    return y
end

CoreFun(y::Matrix{<:Real}) = CoreFun(F; N=size(y,2)) | (; y)

model = CoreFun(rand(2,1000))

sampler = NUTS();
chain = sample(model, sampler, 4_000, progress=false)
```

causes:

```
MethodError: *(::ForwardDiff.Dual{ForwardDiff.Tag{DynamicPPL.DynamicPPLTag, Float64}, Float64, 1}, ::TaylorDiff.TaylorScalar{ForwardDiff.Dual{ForwardDiff.Tag{DynamicPPL.DynamicPPLTag, Float64}, Float64, 1}, 2}) is ambiguous.

Candidates:
  *(x::ForwardDiff.Dual{Tx}, y::Real) where Tx
    @ ForwardDiff C:\Users\Monali\.julia\packages\ForwardDiff\UBbGT\src\dual.jl:148
  *(a::Number, b::TaylorDiff.TaylorScalar)
    @ TaylorDiff C:\Users\Monali\.julia\packages\TaylorDiff\N6Pcf\src\primitive.jl:114

Possible fix, define
  *(::ForwardDiff.Dual{Tx}, ::TaylorDiff.TaylorScalar) where Tx


Stacktrace:
[...]

PR #302, opened 2025-09-19
Copulas.jl — coherent `fit` API aligned with StatsBase/Distributions

Santymax98:
## Summary

This PR introduces a **unified fitting API** and a lightweight model object so copula fitting behaves consistently with the Julia stats ecosystem. The first milestone targets **one-parameter Archimedean families** plus **empirical** copulas.

---

## Highlights

* New `CopulaModel` container (subtypes `StatsBase.StatisticalModel`) returned by `Distributions.fit`.
* Clean dispatcher with sensible defaults and clear errors.
* Rank-based estimators (`:itau`, `:irho`, `:ibeta`) and `:mle` for 1-par Archimedean copulas.
* Empirical fits (`:emp`) with **Beta**, **Bernstein**, and **raw Empirical** copulas; **Empirical EV tail** for extreme-value families.
* StatsBase/Distributions integration: coefficients, vcov, SEs, CIs, information criteria, loglik, etc.
* Polished `show(::CopulaModel)` with LR test and a compact **coefficient table** (parametric) or **empirical summary** (τ, ρ, β).

---

## New types & user API

* `struct CopulaModel{C,TM,TD} <: StatsBase.StatisticalModel`

  * Fields: `copula`, `n`, `ll`, `method`, `vcov`, `converged`, `iterations`, `elapsed_sec`, `method_details`.
* Top-level entry point:

  ```julia
  M = fit(CopulaModel, ClaytonCopula, U; method=:mle)
  M = fit(CopulaModel, ArchimedeanFamily, U; method=:itau)  # one-param Archimedean
  M = fit(CopulaModel, Copulas.Copula,   U; method=:emp)    # generic empirical
  ```
* Accessor:

  ```julia
  C = Copulas.copula(M)   # or M.copula
  ```

---

## Implemented fitting methods

**Archimedean (1-parameter)**

* `:mle` — 1D Brent on the **negative loglikelihood** with open bounds and safe projection.
* `:itau` — invert Kendall’s τ pairwise and average; respects family bounds.
* `:irho` — invert Spearman’s ρ pairwise and average; respects bounds.
* `:ibeta` — invert Blomqvist’s β (multivariate) via bracketing + Brent; robust when bounds are ±∞.

**Empirical (`:emp`)**

* `BetaCopula` (default), `BernsteinCopula` (with `m`), or `EmpiricalCopula` (raw).

Each fit returns `(copula, method_details::NamedTuple)`. The dispatcher wraps this into a `CopulaModel` and augments metadata (e.g., `null_ll`).

---

## StatsBase/Distributions integration

For `CopulaModel`:

* `nobs`, `dof`, `coef`, `coefnames`, `vcov`, `stderror`, `confint`, `deviance`, `aic`, `bic`, `isfitted`.
* `Distributions.loglikelihood(M)` and `Distributions.loglikelihood(copula(M), U)`.

For `Copula`:

* `Distributions.loglikelihood(C, U)` (fast column loop over `logpdf`).

Degrees of freedom:

* Generic `dof(Copula) = length(Distributions.params(C))` with special cases `dof(::Empirical*) = 0`.
  (Parametric specializations can extend this as needed.)

---

## Output / UX

* **Parametric**: prints N, null loglik, loglik, LR χ² test, AIC/BIC, convergence info, and a coefficient table with SE, z, p, CI.
* **Empirical** (`dof==0`): prints an **Empirical summary**:

  * For $d=2$: τ, ρ, β (single values).
  * For $d>2$: τ and ρ **pairwise** (mean/sd/min/max) + **β(mult)**;

---

## Dispatcher rules (current milestone)

* **Archimedean (1-par)**: `:mle | :itau | :irho | :ibeta | :emp`.
* **Other copulas (incl. abstract `Copula`)**: `:emp` only (for now).
  Unknown methods for unsupported families raise a precise `ArgumentError`.
* Defaults: Archimedean→`:mle`.

---

## Examples

```julia
# Parametric Archimedean
C = ClaytonCopula(3, 2.5)
U = rand(C, 250)
M = fit(CopulaModel, ClaytonCopula, U; method=:mle)
cop = Copulas.copula(M)
θ   = StatsBase.coef(M)
V   = StatsBase.vcov(M)

# Empirical (generic)
Mβ = fit(CopulaModel, Copulas.Copula, U; method=:emp)  # BetaCopula by default
```

---

## Breaking / notes

* Dispatcher restricted to Archimedean one-parameter families for parametric methods at this stage.
* Nonparametric families report `dof=0`, hence `AIC==BIC==-2*logLik`.
* Export `copula` if we want unqualified access (`using Copulas: copula`).

---

## TODO (next steps)

* Vcov for `:itau/:irho/:ibeta` (jackknife-obs by default; bootstrap option).
* Extend parametric support to multi-parameter Archimedean (e.g., BB families) and others (Gaussian/T/EV).
* Optional MDE pipeline (Cramér–von Mises / KS) once parametric coverage is complete.

@lrnv Also, I needed to add the rho method for Archimedean copulas. Perhaps you have a better understanding of this field because I had problems when I increased the dimension and the constructors gave errors due to monotonicity. So we should work on that. Let me know how we can work and what your thoughts are on it.

### DIFF
--- a/src/Copula.jl
+++ b/src/Copula.jl
@@ -77,7 +77,7 @@ function γ(C::Copula{d}; nmc::Int=100_000, rng::Random.AbstractRNG=Random.Merse
     m /= nmc
     return (m - a_d) / (b_d - a_d)
 end
-ι
+
 function ι(C::Copula{d}; nmc::Int=100_000, rng::Random.AbstractRNG=Random.MersenneTwister(123)) where {d}
     U = rand(rng, C, nmc)
     s = 0.0


Copulas.jl
```diff
@@ -0,0 +1 @@
+Subproject commit 7c61c1498b7567bc025f8bacb8a7479fb0ec96fd
```
lrnv, line 1: This file should not exist ?

src/utils.jl
```diff
@@ -68,4 +68,25 @@ end
...
+# We keep the simple O(n^2) version for clarity and modest n; contributions welcome to add a fast path.
+
+@inline function _winsorize_tau_vclib(τ::Float64)
+    s = τ < 0 ? -1.0 : 1.0
+    a = abs(τ)
+    a = a < 0.01 ? 0.01 : (a > 0.9 ? 0.9 : a)
+    return s*a
+end
+
+# β̂ multivariate (Hofert–Mächler–McNeil, ec. (7))
```
lrnv, line 80: Can you add the citation correctly ?

src/Copula.jl
```diff
@@ -31,6 +31,18 @@ function τ(C::Copula)
     r = Distributions.expectation(F,C; nsamples=10^4)
     return 4*r-1
 end
+function β(C::Copula{d}) where {d}
```
lrnv, line 34: So this beta does have a different deffinition in univariate and multivariate versions. I feel like we have to correctly define \tau and \rho when d > 2, the current definitions are wrong (and then if you want to itau and irho for some multivaraite models, you need the right definition)

src/Generator/JoeGenerator.jl
```diff
@@ -73,3 +73,43 @@ function τ⁻¹(::Type{T},tau) where T<:JoeGenerator
         return Roots.find_zero(θ -> _joe_tau(θ) - tau, (one(tau),tau*Inf))
     end
 end
+
+function _rho_joe_via_cdf(θ; rtol=1e-7, atol=1e-9, maxevals=10^6)
```
lrnv, line 77: So this rho is valid only for d=2 or for other dimensions as well ? Then the rho_inv fuction below is valid for d=2 only or for other dimensions as well ?

src/Generator/GumbelGenerator.jl
```diff
@@ -91,4 +91,47 @@ function Distributions._logpdf(C::ArchimedeanCopula{2,G}, u) where {G<:GumbelGen
     A = LogExpFunctions.logaddexp(θ * lx₁, θ * lx₂)
     B = exp(A/θ)
     return - B + x₁ + x₂ + (θ-1) * (lx₁ + lx₂) + A/θ - 2A + log(B + θ - 1)
+end
+
+function _rho_gumbel_via_cdf(θ; rtol=1e-7, atol=1e-9, maxevals=10^6)
```
lrnv, line 96: I feel like this is only bivariate ?

src/Generator/GumbelBarnettGenerator.jl
```diff
@@ -148,4 +149,49 @@ function τ⁻¹(::Type{T}, tau) where T<:GumbelBarnettGenerator
...
+
+# Edge utilities and robust bracketing
+_GB_EPSA = 1e-12                 # separa de 0
+_GB_EPSB = 1e-12                 # separa de 1
+c_GB_TOLV = 1e-12                 # tolerancia en valor
+
+# Internal grids to rescue bracketing if there are numerical problems
+_GB_GRID_A = (1e-12, 1e-10, 1e-8, 1e-6, 1e-4, 1e-3, 5e-3, 1e-2, 5e-2)
+_GB_GRID_B = (1 - 1e-12, 1 - 1e-10, 1 - 1e-8, 1 - 1e-6, 1 - 1e-4, 0.999, 0.99, 0.95, 0.9)
+function ρs_GB(θ::Real)
```
lrnv, line 162: Again, clarify what happens when d > 2. Also for the \tau and \tau_inv functions, same issue !!!

src/Generator/FrankGenerator.jl
```diff
@@ -73,3 +73,32 @@ function τ⁻¹(::Type{T},tau) where T<:FrankGenerator
         return s*Roots.fzero(x -> _frank_tau(x)-v, 0, Inf)
     end
 end
+
+function ρ(G::FrankGenerator)
```
lrnv, line 77: Again, clarify for d > 2, same idea for \tau

src/Generator/ClaytonGenerator.jl
```diff
@@ -78,4 +78,44 @@ function Distributions._logpdf(C::ClaytonCopula{d,TG}, u) where {d,TG<:ClaytonGe
...
+    θ0 = clamp(θ0, -1 + sqrt(eps(Float64)), 1e6)
+    θ1 = θ0 + (_ρ > 0 ? 0.25 : -0.25)        # segunda semilla hacia el lado correcto
+
+    f(θ) = ρ(ClaytonGenerator(θ)) - _ρ
+    # Secante con dos semillas; no requiere bracketing
+    θ = Roots.find_zero(f, (θ0, θ1), Roots.Order2(); xatol=atol)
+    return θ
+end
+
+ρ⁻¹(::Type{ClaytonCopula}, ρ̂::Real; kwargs...) = ρ⁻¹(ClaytonGenerator, ρ̂; kwargs...)
```
lrnv, line 121: So this does not depend on the dimension at all ??? Weird

src/Fitting/Stats.jl
```diff
@@ -0,0 +1,115 @@
...
+# Conventions:
+# - Rows = variables (p), columns = samples (n)
+# - This file exposes the StatsBase/Distributions interface for CopulaModel
+
+# ---------- 1) Utils ----------
+@inline _as_pxn(C::Copulas.Copula, U::AbstractMatrix) = (size(U,1) == length(C)) ? U : permutedims(U)
+@inline _as_pxn(p::Integer,         U::AbstractMatrix) = (size(U,1) == p)        ? U : permutedims(U)
+
+# ---------- 2) Log-likelihood (Distributions) ----------
+function Distributions.loglikelihood(C::Copulas.Copula, U::AbstractMatrix{<:Real})
```
lrnv, line 11: Cant this be defined easily as `sum(Distributions.logpdf(C, U)` ? Isnt it alreayd defined for us in Distributions.jl ??

src/Fitting/Stats.jl
```diff
@@ -0,0 +1,115 @@
+# =========================== src/Fitting/Stats.jl ===========================
+# Conventions:
+# - Rows = variables (p), columns = samples (n)
+# - This file exposes the StatsBase/Distributions interface for CopulaModel
+
+# ---------- 1) Utils ----------
+@inline _as_pxn(C::Copulas.Copula, U::AbstractMatrix) = (size(U,1) == length(C)) ? U : permutedims(U)
+@inline _as_pxn(p::Integer,         U::AbstractMatrix) = (size(U,1) == p)        ? U : permutedims(U)
```
lrnv, line 8: Please STOP with this function ^^ 

The standard is that data HAS to be passed in (d,n) size, point. If the user pass it as (n,d) then screw them they get a n-dimensional copula, we do not help with it.

src/Fitting/CopulaModel.jl
```diff
@@ -0,0 +1,276 @@
...
+    function CopulaModel(c::C, n::Integer, ll::Real, method::Symbol;
+                         vcov=nothing, converged=true, iterations=0, elapsed_sec=NaN,
+                         method_details=NamedTuple()) where {C}
+        return new{C, typeof(vcov), typeof(method_details)}(
+            c, n, float(ll), method, vcov, converged, iterations, float(elapsed_sec), method_details
+        )
+    end
+end
+
+copula(M::CopulaModel) = M.copula
```
lrnv, line 22: Please remove this

src/Fitting/CopulaModel.jl
```diff
@@ -0,0 +1,276 @@
...
+                                * "Use method=:emp or set a specific Archimedean family."))
+        end
+    end
+end
+
+# --------------------------- Distributions.fit -------------------------------
+
+function Distributions.fit(::Type{CopulaModel}, ::Type{T}, U::AbstractMatrix; method::Symbol=:auto, kwargs...) where {T<:Copula}
+    m        = _select_method(T, method)
+    C, meta  = _dispatch_fit(T, U, m; kwargs...)
```
lrnv, line 131: Here pass `Val{m}()` instead of m

src/Fitting/CopulaModel.jl
```diff
@@ -0,0 +1,276 @@
...
+            throw(ArgumentError("method=$(method) is not implemented for $(T). "
+                                * "Use method=:emp or set a specific Archimedean family."))
+        end
+    end
+end
+
+# --------------------------- Distributions.fit -------------------------------
+
+function Distributions.fit(::Type{CopulaModel}, ::Type{T}, U::AbstractMatrix; method::Symbol=:auto, kwargs...) where {T<:Copula}
+    m        = _select_method(T, method)
```
lrnv, line 130: Here pass `Val{m}()` instead of m

src/Fitting/CopulaModel.jl
```diff
@@ -0,0 +1,276 @@
...
+
+    _print_coeftable(io, M)  # (sin cambios)
+end
+
+_default_method(::Type{<:ArchimedeanCopula})  = :mle
+# ====================== Dispatch ======================
+@inline _select_method(::Type{T}, method::Symbol) where {T<:Copula} =
+    (method === :auto) ? _default_method(T) : method
+
+@inline function _dispatch_fit(::Type{T}, U::AbstractMatrix, m::Symbol; kwargs...) where {T<:Copula}
```
lrnv, line 99: This looks like you are reinventing dispatch... Please dont ;) Simply pass the method as `Val{m}()`, and then you can write functions like : 

```julia
_fit(::ArchimedeanCopula, ::AbstractMatrix, ::Val{:mle}) = ...
_fit(::ArchimedeanCopula, ::AbstractMatrix, ::Val{:itau}) = ...
_fit(::ArchimedeanCopula, ::AbstractMatrix, ::Val{:irho}) = ...
```

src/Fitting/ArchimedeanFit.jl
```diff
@@ -0,0 +1,363 @@
...
+        _vcov_jackknife_obs(CT, U; estimator=:ibeta, epsβ=epsβ) :
+        error("vcov_method no soportado")
+    return C, (; estimator=:ibeta, epsβ=epsβ, vcov=V, vcov_details=vmeta)
+end
+# ============================= :MLE =======================================
+# Seed
+function _seed1(::Type{CT}, d::Integer, Up::AbstractMatrix, start) where {CT<:ArchimedeanCopula}
+    GT = generatorof(CT)
+    if start === :itau
+        return Distributions.params(_fit_itau_1par(CT, Up))[end]
```
lrnv, line 204: Here too, you are reinventing dispatch... Simply define 

```julia
_fit(T::Type{Archimedean}, u::Abstractmatrix, ::Val{:default}) = _fit(T, u, ::Val{:mle}) # to set the default. 
_fit(::Type{Archimedean}, ::Abstractmatrix, ::Val{:itau}) = ...  # with the real estimator. 
_fit(::Type{Archimedean}, ::Abstractmatrix, ::Val{:irho}) = 
_fit(::Type{Archimedean}, ::Abstractmatrix, ::Val{:ibeta})
_fit(::Type{Archimedean}, ::Abstractmatrix, ::Val{:mle})
```
and the same for other models, keep one function name and overlad its methods

src/Fitting/ArchimedeanFit.jl
```diff
@@ -0,0 +1,363 @@
...
+    return C, (; estimator=:irho, epsρ=epsρ, vcov=V, vcov_details=vmeta)
+end
+
+# ============================= :IBETA =======================================
+function _fit_ibeta_1par(::Type{CT}, U::AbstractMatrix;
+                         epsβ::Real=1e-10, max_expand::Int=20) where {CT<:ArchimedeanCopula}
+    # data
+    d = (size(U,1) ≥ 2) ? size(U,1) : size(U,2)
+    d ≥ 2 || throw(ArgumentError("fit_ibeta(Archimedean) requires d≥2."))
+    Up = _as_pxn(d, U)
```
lrnv, line 139: Please ditch the as_pxn behavior

lrnv (changes requested):
I have left a few comments right and left, but there a two most important ones: 

1) You should use multiple dispatch with only one function name (I propose `_fit()`) instead of defining `_fit_archi_1par_ibeta_only_if_it_rains()` and dispatch by hand to it. To dispatch on symbols, use Val{}. That will make the internals much more clear to see what is hapening. 

2) If you want a static way to filter some types with respect to others, for eample to filter only univariate generators: 

```julia
abstract type UnivaraiteGenerator<:Generator end
abstract type BivariateGenerator<:Generator end

# and then 
struct ClaytonGenerator<:UnivaraiteGenerator
```

3) I am NOT SURE about the current definition of \tau and \rho when d>2, I think we should investigate about these.

src/Fitting/CopulaModel.jl
```diff
@@ -0,0 +1,276 @@
...
+        end
+    end
+end
+
+# --------------------------- Distributions.fit -------------------------------
+
+function Distributions.fit(::Type{CopulaModel}, ::Type{T}, U::AbstractMatrix; method::Symbol=:auto, kwargs...) where {T<:Copula}
+    m        = _select_method(T, method)
+    C, meta  = _dispatch_fit(T, U, m; kwargs...)
+    Up       = _as_pxn(C, U)
```
lrnv, line 132: Ditch as_pxn

Santymax98:
ok, I'm waiting

lrnv:
Okay I have took a first shot at defining what we want. If you can read carefully the diff of my commit that would be great. I have to go sleep now (its late in europe) I'll continue tomorow. 

Basically we need to define more carefully the `Distributions.params` functions (to get parameter names), and maybe a `param_bounds` that goes with it for each model ? 

Also I do not agree that the defautl behavior shoudl compute thousands of things (jacknife and others), this should be on-demand and not by default, so i removed a few things but I basically broke everything for the moment

Santymax98:
Following the recommendations, I tried to improve what I had a little and added Sklardist fit to see what you think...

lrnv:
Santiago I took the time to remove the conditionals and move to proper dispatching for the empirical copulas, and your removed everything to restore your previous state ? That is not cool, please be carefull of what you commit

lrnv:
Alsào stop writing helper function used only once. If used only once, it does not need to be a function.

lrnv:
So, I did the following: 

- I made the extra computations to be ON by default when calling `fit(CopulaModel, T, ...)` but OFF by defautl when calling `fit(T, ...)`. I think its a good compromise
- I have redone the univaraite generator bindings, even if it is not defined yet it will come. 
- Moved everything to a fitting.jl file. `_fit()` methods and other things specific to models will have to be moved to their repsective files before merging but for the moment its easier like that. 
- Refactored utility functions in utils.jl to improve performance and readability.
- Enforced proper dispatching, no more if statements on the methods... in partiuclar for the empirical models, please do not revert this once more... 
- Contained the show function to one method, please do not expand it to thousands of one-use functions. 
- On archimedeans fits, I simplified the code it was very cluttered with excessive care all around... 

Please do not re-propose things that I removed this time. Maybe we need to look at tests cases now,

Santymax98:
I fixed some tyny bugs ^^ and noticed your question on line 127. Thanks for reporting it. 😉

I fully agree with your refactoring, and my (somewhat tentative) opinion is:

* `coef(M)` / `coefnames(M)` should expose only the copula parameters, to be consistent with `vcov`/`confint` (which we currently compute only for the copula).
* AIC/BIC should account for all parameters (copula + margins) using `dof(::SklarDist)`.
* The LR vs. independence test should use only the copula's degrees of freedom (standard IFM practice).
* In `show`, kept the **\[Copula]** and **\[Marginals]** sections separate.

If/when we add a joint `vcov` (copula + margins), we could introduce a function like `fullcoef(M)` (or `coef(M; include_margins=true)`) to present everything consistently.

It would be interesting to adapt the semantics if you prefer a different one; but the idea I had at the time was that we just wanted to keep the inference consistent. I don't know if I'm wrong haha ​​I feel that vcov is very important in our interface...

Tell me what you think :)

lrnv:
Okay I have made all univariate archiemdean fits work with :mle, :itau, :irho, and :ibeta for some of them (not all). 

I will now look at other models. I feel like we need some tests ?

Santymax98:
Do you think this is what we're looking for? If so, I think we should change some methods in Stats.

For example:

`StatsBase.coef(M) = collect(values(Distributions.params(copula_of(M))))`

`StatsBase.coefnames(M) = string.(keys(Distributions.params(copula_of(M))))`

`StatsBase.dof(C::Copulas.Copula) = length(values(Distributions.params(C)))`

To reconstruct the copula, we could create something like:

`from_params(::Type{ClaytonCopula}, d, p) = ClaytonCopula(d, p.θ)`

This enables tests like: `from_params(typeof(C), length(C), params(C)) ≈ C` and opens the door to a generic MLE later.

lrnv:
@Santymax98 I have tried to implement the upper interface for every model. I have pushed everything, but maybei did mistakes, now we need to check that every fit will go through. I have added some tests, but this will neot be enough, let me add more.

Santymax98:
Hi Oskar, I tried to do it with TCopula and Gaussian Copula, and the general one succeeded... (I'm referring to the mle you implemented) just now I was going to send the commit, but if I do, is there a rebase? I'm bad at using Git.

Santymax98:
well, I'll write it here so you can see it.
## Gaussian Copula
```julia
function Distributions.params(C::GaussianCopula)
    Σ = C.Σ; d = size(Σ,1)
    return (; (Symbol("ρ_$(i)$(j)") => Σ[i,j] for i in 1:d-1 for j in i+1:d)...)
end
_example(::Type{GaussianCopula}, d::Int) = GaussianCopula(Matrix(LinearAlgebra.I, d, d) .+ 0.2 .* (ones(d, d) .- Matrix(LinearAlgebra.I, d, d)))

function _unbound_params(::Type{GaussianCopula}, d::Int, θ::NamedTuple)
    Σ = _Σ_from_named(d, θ)
    L = LinearAlgebra.cholesky(LinearAlgebra.Symmetric(Σ), check=false).L
    δ = log.(LinearAlgebra.diag(L))
    ℓ = Vector{eltype(Σ)}(undef, d*(d-1)÷2); k=1
    @inbounds for i in 2:d, j in 1:i-1
        ℓ[k] = L[i,j]; k+=1
    end
    return vcat(δ, ℓ)
end

function _rebound_params(::Type{GaussianCopula}, d::Int, α::AbstractVector{T}) where {T}
    L = Matrix{T}(LinearAlgebra.I, d, d)
    @inbounds begin
        for i in 1:d; L[i,i] = exp(α[i]); end
        k=1
        for i in 2:d, j in 1:i-1
            L[i,j] = α[d+k]; k+=1
        end
    end
    S  = L*L'
    dvec = sqrt.(LinearAlgebra.diag(S))
    Dinv = LinearAlgebra.Diagonal(inv.(dvec))
    Σ = Dinv * S * Dinv
    Σ = (Σ + Σ')/2   # force simetric matrix ... 
    return (; Σ = Σ)
end
#only for flexibility in our fit... other option could be return CT(θhat.Σ), meta in generic mle fit...
GaussianCopula(d::Int, Σ::AbstractMatrix) = GaussianCopula(Σ) 
```

# TCopula

```julia
function Distributions.params(C::TCopula{d,df,MT}) where {d,df,MT}
    Σ = C.Σ; n = size(Σ,1)
    return (; ν = df, (Symbol("ρ_$(i)$(j)") => Σ[i,j] for i in 1:n-1 for j in i+1:n)...)
end
_example(::Type{TCopula}, d::Int) = TCopula(5.0, Matrix(LinearAlgebra.I, d, d) .+ 0.2 .* (ones(d, d) .- Matrix(LinearAlgebra.I, d, d)))
function _unbound_params(::Type{TCopula}, d::Int, θ::NamedTuple)
    Σ = _Σ_from_named(d, θ)
    L = LinearAlgebra.cholesky(LinearAlgebra.Symmetric(Σ), check=false).L
    δ = log.(LinearAlgebra.diag(L))
    ℓ = [L[i,j] for i in 2:d for j in 1:i-1]
    return vcat(log(θ.ν), δ, ℓ)
end

function _rebound_params(::Type{TCopula}, d::Int, α::AbstractVector{T}) where {T}
    ν = exp(α[1])
    L = Matrix{T}(LinearAlgebra.I, d, d)
    for i in 1:d; L[i,i] = exp(α[1+i]); end
    k = 1
    for i in 2:d, j in 1:i-1
        L[i,j] = α[d+1+k]; k+=1
    end
    S  = L*L'
    dvec = sqrt.(LinearAlgebra.diag(S))
    Σ = LinearAlgebra.Diagonal(inv.(dvec)) * S * LinearAlgebra.Diagonal(inv.(dvec))
    Σ = (Σ + Σ')/2   # fuerza simetría
    return (; ν = ν, Σ = Σ)
end

TCopula(d::Int, ν::Real, Σ::AbstractMatrix) = TCopula(ν, Σ)
```

In èllipticalcopulas.jl``i add

```julia
#this function is for test your idea 
@inline function _Σ_from_named(d::Int, θ::NamedTuple)
    Tρ = eltype(values(θ))
    Σ  = Matrix{Tρ}(LinearAlgebra.I, d, d)
    @inbounds for i in 1:d-1, j in i+1:d
        ρ = θ[Symbol("ρ_$(i)$(j)")]
        Σ[i,j] = Σ[j,i] = ρ
    end
    return Σ
end
```

for finished... I modified your prototype like this because there are problems with the automatic differentiation of the T, remember?

```julia
function _fit(CT::Type{<:Copulas.Copula}, U, ::Val{:mle})
    d   = size(U,1)
    θ₀  = Distributions.params(_example(CT, d))
    α₀  = _unbound_params(CT, d, θ₀)
    loss(α) = -Distributions.loglikelihood(CT(d, _rebound_params(CT, d, α)...), U)

    res = try
        Optim.optimize(loss, α₀, Optim.LBFGS(); autodiff=:forward)
    catch err
        @warn "LBFGS with AD failed ($err), retrying with NelderMead"
        Optim.optimize(loss, α₀, Optim.NelderMead())
    end

    θhat = _rebound_params(CT, d, Optim.minimizer(res))
    return CT(d, θhat...),
           (; θ̂=θhat, optimizer=Optim.method(res), converged=Optim.converged(res),
              iterations=Optim.iterations(res))
end
```

lrnv:
Yes please do that, and cleanup what i did in those two files, your verison is way better.

Santymax98:
Thaks í will this

Santymax98:
@lrnv Do you know any other elliptical copula different from these two?

lrnv:
No i do not. I am sure there are others, but hese two are really the most used ones. 

Basically from any distribution on R_+ you could define an alliptical copula (where if you want the gaussian you'll take the positive part of a univariate Normal), but I never had the need to implement more general things that we we have here.

Santymax98:
Okay, to avoid overshooting, tell me where you'd like me to help you. While you make the necessary changes for the first shot.

lrnv:
I'm stopping now for the night, comming back tomorow. So take the models you want in the fittingtest and try to have the fit work for as much methods as you can ;)

Santymax98:
Okay, I solved the problem with ibeta for Archimedeans. In general, I think the problem is that it was sometimes very flat, and the method couldn't find a change of sign. On the other hand, I noticed that if $\rho$ isn't implemented, the cost will be quite high...

I think EVCopulas work with the generic mle that has been implemented.

lrnv:
> On the other hand, I noticed that if $\rho$ and $\tau$ aren't implemented, the cost of evaluating these functions with numerical integration is very high.

Maybe the default itau/irho fitting functions should check for existence of specific methods (like we do in the tests) and warn the user that it might take some time if there is none.

lrnv:
It looks like the unconstraint parameters of the Gaussian copula dont have the right number of values, this is weird. Maybe it is because the transfrom did not took into account that the matrix is a corelation matrix (and thus diagonal === 1) ?

Santymax98:
I implemented some things for extreme value copulas and tried to replicate your methodology for a single parameter in `univariategenerator` but applied it to `Tail2`. However, I ran into some issues, and I don't think the methods are being redirected. Could you review it to see where the problem lies?

lrnv:
> It doesn't work with Archimedeans either... I don't know if I'm wrong, but when I do the fit... it redirects me to the generic ones and not to the specific ones that have been written.

That is an issue. Could be checked out by simply adding @show "version 1" and @show "version 2" to both. But its true that dispatching on Types as we are doing is always a bit finicky. For example, `Type{ClaytonCopula}` and `Type{<:ClaytonCopula}` are really not the same thing.

Santymax98:
Yes, please, at least one and i could change the rest myself

lrnv:
So you are true that it is not edispatching to the generic one. For example for the mle: 

```julia
julia> ClaytonCopula isa Type{<:ArchimedeanCopula{d, GT} where d} where GT<:Copulas.UnivariateGenerator
false

julia> ClaytonCopula isa Type{<:ArchimedeanCopula{d, <:Copulas.UnivariateGenerator} where d}
true

julia> 
```

These two type statements are NOT equivalent, the second is equivalent to `Type{<:ArchimedeanCopula{d, GT} where {d, GT<:Copulas.UnivariateGenerator}}` i think. So the current version of the method has to be changed, let me do that in a commit then you can replicate the change elsewhere.

lrnv:
I thkn i had left a `where GT` at the end of the line that should not be there...

Santymax98:
well, now yes!!!!!!!
```julia
julia> fit(ClaytonCopula, U)
"Running the MLE routine from the Archimedean Univaraite implementation" = "Running the MLE routine from the Archimedean Univaraite implementation"
ClaytonCopula{2, Float64}(θ = 3.0819534068408982,)

julia> 
```

lrnv (commented):
No problem, keep the direction you had I will continue writting tests.

lrnv (commented):
Still problematic for me: BC2Copula, tEVCopula, MOCopula, AsymLogCopula, AsymGalambosCopula, AsymMixedCopula, ArchimaxCopula{2, Copulas.GumbelGenerator, Copulas.MixedTail}, BB4Copula, and BB5Copula

Santymax98:
Okay... I tried hard but it was the only way to solve it... the types were quite complicated in this case...

lrnv (commented):
I think the issue comes from the fact that `BB4Copula = ArchimaxCopula{2, XXGen{T}, YYTail{T}}` with the same T... This is adesign flaw. 

I feel like in the PR we are specializing methods that we could figure out at higher levels... but its not easy to deal with the types indeed, maybe in a next PR about simplification. Even the constructors of archimedeans could be written only once... But for that we need the generators and tails to have coherent type parametrisation all across, which is not easy to do.

Santymax98:
Do you want me to implement a version of the tail dependence? Cortail style?

Santymax98:
Yes, I will do it later. Right now I am at a conference, haha, and I don't have access, but I have it ready to implement it and complete what is missing.

Santymax98:
It's okay, I just improved a little thing that was missing in the references 👍

lrnv:
its called the weak tail dependence function, i dont really know a reference. Maybe in Nelsen 2006, In Joe 2014 or in Durante 2015 ?

lrnv:
There is also charpentier's blog with a part 1/2/3 articles : https://freakonometrics.hypotheses.org/?s=copulas+and+tail+dependence

Santymax98:
Hi @lrnv , I think we should modify the entropy letter, perhaps to the capital letter \Eta or simply the normal letter H. This is because when I was working with the weak tail coefficient chi, I noticed that in the literature the letter \eta is used for the parameter that controls the decay in the weak tail.

lrnv:
@Santymax98 I think I will merge this now. Reminding are : 

1) The vcov things to be able to output in the show() function of CopulaModel. 
2) Improved testing for codecov. 
3) Improved documentation. 

I expect the two first points to become two different pull-requests from you, and I will handle the documentation myself. I will not publish a new versions before these three points are covered, but since they are rather "peripherical" to the central fitting things, they can be in three other pull-requests. 

GG on this one !

lrnv:
If you have issues with forwarddiff that look like #301 then a fix is going to come but not right now, waiting on TaylorDiff.jl's guys to fix it. Anyway let's discuss all that in a new PR, this one was becommin soooooo big it had to stop ^^

lrnv:
This is because in BB1Generator, there is 

```julia
function Distributions._logpdf(C::ArchimedeanCopula{2,G}, u) where {G<:BB1Generator}
    T = promote_type(Float64, eltype(u)) # <<<<<<<<<<<<<<<<<<<<<<<<<<-------- Here is your issue. 
    (0.0 < u[1] ≤ 1.0 && 0.0 < u[2] ≤ 1.0) || return T(-Inf)

    θ, δ = C.G.θ, C.G.δ
    a    = inv(δ)
    w1   = expm1(-θ*log(u[1]))               # u1^{-θ} - 1
    w2   = expm1(-θ*log(u[2]))               # u2^{-θ} - 1
    s    = w1^δ + w2^δ
    sa   = s^a

    # log c(u,v) = log ψ''(s) + log (ψ^{-1})'(u) + log (ψ^{-1})'(v)
    logc = log(δ*θ) +                         # (= log(a/θ) + 2log(δθ))
           (a - 2)*log(s) +
           (-inv(θ) - 2)*log1p(sa) +
           log((1 + a/θ)*sa - (a - 1)) +
           (δ - 1)*(log(w1) + log(w2)) +
           (-θ - 1)*(log(u[1]) + log(u[2]))

    return T(logc)
end
```

Where the two first lines should relaly be 
```julia
function Distributions._logpdf(C::ArchimedeanCopula{2,BB1Generator{TF}}, u) where {TF}
    T = promote_type(TF, eltype(u)) # <<<<<<<<<<<<<<<<<<<<<<<<<<-------- Here is your issue. 
```

or something like that. The same error might also appear in other generators ;)

PR #310, opened 2025-10-02
Vcov fitting

Santymax98:
I tried to implement vcov for the models, letting the user decide whether they want vcov or not. I also added the function to print the dependency metrics in Sklar Dist (again, the decision to print it or not is up to the user).


Santymax98:
I think I made a mistake in the branch....

lrnv:
Hum... Looks like you have everything that is now on master here. Can you rebase this ont he current master so that we only heav eht echanges related to the vcov ?

PR #340, opened 2025-10-18
[Internals] Internalize integration routines

lrnv:
To be able to ditch HCubtature.jl and thus StaticArrays.jl, which increase load time. Hopefully without performance losses ? Not sure yet. 

This work but its awfully slow, but maybe we can make it faster there are probably inneficinecies all around this implementation we"'ll have to rpoof-èread it. 

Not sure I want to merge it.


lrnv:
wont merge

PR #355, opened 2026-04-14
Update Optim to version 2

FriesischScott:
This PR updates the compat entry for *Optim* to 2.0 and makes the necessary changes required by the new major release. 

Due to a change of behaviour in *ForwardDiff* I've also removed the compat for version `1`. This issue can already be seen on `main` right now. From the `1.0` release of  *ForwardDiff*  `Dual(1.0, 1.0) > 1.0` is `true`,  which was `false` before, because it includes the partial in the comparison. This breaks some of the fitting where the objective includes bound checks.

I had to change the fitting of the `ArchimedianCopulas` to use `IPNewton` and start from the midpoint of `lo` and `hi` to make the tests pass. Overall i think `FminBox` is being phased out in favour of `IPNewton`.

Similarly, for the `ExtremeValueCopula` I had to clamp the starting value to `floatmax` when the upper bound is `inf`.

Finally, I removed some type inconsistencies from the `_θ_bounds` functions so they always return floats. There are some white space changes here and there because my editor removed all trailing spaces in the files I touched.

This PR replaces #348.

We're hoping to merge some code that requires *Optim* `2` but are currently blocked by *Copulas*. It would be fantastic if you could tag a new release that is compatible with *Optim* `2.0`.


devmotion:
> This breaks some of the fitting where the objective includes bound checks.

Instead of dropping support for ForwardDiff@1 (which would mean that many users wouldn't even being able to install the latest versions of Copulas in their environments), it would be better to fix the bound checks.

FriesischScott:
> > This breaks some of the fitting where the objective includes bound checks.
> 
> Instead of dropping support for ForwardDiff@1 (which would mean that many users wouldn't even being able to install the latest versions of Copulas in their environments), it would be better to fix the bound checks.

I am still hoping I can workout a solution.

FriesischScott:
Turns out the reason wasn't actually *ForwardDiff* but the initial guess for the optimization when they are too close to or right on the variable bounds. I'm now restricting `Inf` bounds to `1e16` and zero to `1e-16`. Going higher or lower quickly leads to numerical issues. The optimization itself of course still uses the correct bounds.

Thanks for the push @devmotion!

lrnv:
@FriesischScott Thanks a lot for all this, what is then the current status of tis PR ? Should i review it and merge or you are still working on it ? 

Edit: Having troubles on master which this might fix, so merging and then fixing on master before releasing a new version.

PR #367, opened 2026-05-29
Nested Archimedean copulas

thisiscam:
Adds support for hierarchical Archimedean copulas of arbitrary tree depth with per-subtree generator choices. Density via Faà di Bruno over the tree (Bell-polynomial recursion + Cauchy product per node); closed-form CDF.

Reuses upstream `ϕ⁽ᵏ⁾`, `taylor()`, and the existing generator interface throughout — no parallel derivative machinery. Tests check the tree-density against an independent nested-`ForwardDiff` reference and against frozen CSVs from a JAX implementation.

Three coordination questions before un-drafting:

1. `src/nested/FrankTaylorGenerator.jl` is 4 lines adding `ϕ` / `ϕ⁻¹` of `FrankGenerator` on `::Taylor1` (works around `log1mexp` not having a Taylor1 rule). Happy to split as a tiny `main`-targeted PR first and rebase.
2. `src/CensoredLikelihood.jl` currently has our per-variable censored likelihood. Once #345 lands, the natural form is the decomposition via `condition` / `subsetdims`. Prefer (a) wait for #345 and rebase onto it, (b) drop censoring from this PR and follow up separately, or (c) keep here for now and remove on rebase?
3. `condition` / `subsetdims` aren't specialised for `NestedArchimedeanCopula` yet — the generic AD-on-`_cdf` fallback works (we provide a closed-form `_cdf`) but is slow. Want a tree-walking specialisation in this PR or as a follow-up?

Reference: Yang & Li, "Archimedean Copula Inference via Taylor-Mode AD," arXiv:2605.23134 (Algorithm 1 = the nested-AC density).


lrnv:
1. I understand why you want to do that, but our interface should already cover the case fully through `ϕ, ϕ⁻¹, ϕ⁽¹⁾, ϕ⁻¹⁽¹⁾, ϕ⁽ᵏ⁾` and `ϕ⁽ᵏ⁾⁻¹`. Do you really need `ϕ⁻¹⁽ᵏ⁾` for k > 1 ?
	(a) You do need `ϕ⁻¹⁽ᵏ⁾` for k > 1: Then maybe we should provide a generic interface so that it is available to all generators, and can be overwritten, as we did for others. 
	(b) You do not need it: Then please do not use Taylor1 directly and pass through our generics: this will make the Frank one go through without a specific fix. This should avoid other issues as the ones you might have with very-dependent gumbels, or other things like that. 

2. The other PR landed alreayd on master, but i still prefer (b) make it a separate PR, as this clearly is a separate feature

3. If possible, yes, it would be better to support the whole interface. Also, the fit() function ? At least maximum likelihood should be possible if the tree and generators are provided, since we have a density.

thisiscam:
> 1. I understand why you want to do that, but our interface should already cover the case fully through `ϕ, ϕ⁻¹, ϕ⁽¹⁾, ϕ⁻¹⁽¹⁾, ϕ⁽ᵏ⁾` and `ϕ⁽ᵏ⁾⁻¹`. Do you really need `ϕ⁻¹⁽ᵏ⁾` for k > 1 ?
>        (a) You do need `ϕ⁻¹⁽ᵏ⁾` for k > 1: Then maybe we should provide a generic interface so that it is available to all generators, and can be overwritten, as we did for others.

Yes, we'd need $\textrm{taylor1}(\phi_{\text{outer}}^{-1} \circ \phi_{\text{inner}}, d) $. 
A more robust way to fix this, perhaps, is to allow registering custom ops in TaylorSeries.jl, since the op that misses treatment is `log1mexp`. 
Let me know what you think?

lrnv:
So you do need `ϕ⁻¹⁽ᵏ⁾`, k > 1: the higher derivatives of the inverse of the generator. The right thing to do would then be to add the taylor1() generic version of this function to the list of potentially-overwrittable functions in Generator.jl, and then maybe try to fill out its value for the few generators where it is known ? Also in the docs on the generator page there's a mention of the list of potentially-overwritable functions, where it hsoul dbe added. 

Then, you code should not use the taylor1() function but rely on the new `ϕ⁻¹⁽ᵏ⁾`, function, so that if a generator has a faster version (or a version that handles better extreme cases maybe) it'll be used directly. 

---------------

Regarding `log1mexp`, the right thing to do long-term is to propose a package extension to TaylorSeries.jl that handles LogExpFunctions.jl. I already mentioned it [there](https://github.com/JuliaDiff/TaylorSeries.jl/issues/389#issuecomment-3378312599) a while ago but never got the time to do it. 

Adding the two methods you proposed is fine as a temporary measure. 

I feel like those methods belong to the Frank generator file and not really their own file, maybe you could move them there ? 

--------------

Can you please add some examples of nested archimedean copulas to the GenericTests.jl file ? 

--------------

src/nested/FrankTaylorGenerator.jl
```diff
@@ -0,0 +1,33 @@
+# =============================================================================
```
lrnv: The content of this file should be moved to the file of the frank generator

src/nested/NestedArchimedeanDensity.jl
```diff
@@ -0,0 +1,200 @@
...
+    children::Vector{_NestedNode}
+end
+
+# Taylor coefficients [h'(t₀)/1!, …, h⁽ᵈ⁾(t₀)/d!] of the change of variables
+# h = ϕ⁻¹_outer ∘ ϕ_inner at t₀, to order d. This is the inner-to-outer link in
+# the Faà di Bruno recursion for a child sub-tree.
+function _composition_taylor(outer::Generator, inner::Generator, t₀::T, d::Int) where {T}
+    coefs = taylor(x -> ϕ⁻¹(outer, ϕ(inner, x)), t₀, d)
+    return T[coefs[k+1] for k in 1:d]
+end
```
lrnv: So, the derivatives of `ϕ⁻¹(outer, ϕ(inner, x))` should IMHO be obtained by manually running the faa-di-bruno formula leveraging the methods we have for the derivatives of these two functions : computing these derivatives is sometimes challenging in term of precision, and the direct taylor() version does not use the potentially overwritten version of those.

src/nested/NestedConditioning.jl
```diff
@@ -0,0 +1,238 @@
...
+    surv = sort!(vcat(newleaf, [j for ch in newkids for j in _kid_dims(ch)]))
+    return NestedArchimedeanCopula{length(surv), typeof(c.G)}(c.G, newleaf, newkids, surv)
+end
+
+# GLOBAL dims covered by a pruned child entry.
+_kid_dims(ch::Tuple) = ch[2]
+_kid_dims(ch::NestedArchimedeanCopula) = ch.dims
+
+function SubsetCopula(C::NestedArchimedeanCopula{d, TG}, dims::NTuple{p, Int}) where {d, TG, p}
+    # `subsetdims` already short-circuits p==1 (Uniform) and p==d (C unchanged),
```
lrnv: subsetdims indeed short-circuits p==1 (Uniform) and `dims == 1:d` (unchanged) but not every `p==d` (reordering)

src/nested/NestedArchimedeanDensity.jl
thisiscam: I deliberatedly chose the current version because 1) it's simple and clear on paper 2) NAC density most likely requires BigFloat in most cases, due to the convolution-of-the-taylor-polynomials step requiring a wide range of precision.  
That said, a more robust version we had in our preprint and also the JAX implementation was based on implicit differentiation (in Appendix A.4), which avoids the high-order derivative of the inverse. 

Do we want that implemented?

lrnv: No, its better to use the higher-order derivatives of the inverses as it allows implementers of generators to specify them if they want. Can you leverage the implementation of FdB/Bell that you already have ?

thisiscam: **1.** I tried FdB method and found it is broken because of catastrophic cancellation at the FdB summation. On a Frank/Frank example, the derivative of the nesting becomes completely incorrect, even after feeding in highly-accurate high-order derivatives using BigFloat-512.
  **2.** The direct jet and the implicit solver are both fine and equally accurate. Max rel. 
  error of the edge-link coefficients vs a 2048-bit reference:
  
  ```
    precision | direct    implicit   fdb
    Float64   | 2.3e-12   2.9e-12    1.2e+00
    64-bit    | 9.4e-16   4.2e-16    2.0e-04
    96-bit    | 2.1e-25   7.7e-25    1.3e-13
  ```

This is on Frank/Frank. From my experience with the JAX package, certain nesting $\phi_1^{-1}\circ\phi_2$  has mathematical cancellations so that the direct method will also fail numerically while the implicit method will stay accurate since it completely avoids calculating the high-order derivative of the inverse. 

  **3.** IMHO the productive fix for this class of numerical issue is to let the user override the composition's Taylor calculation directly (`composition_taylor`). Then, per generator-pair, they can pick the naive taylor method, the implicit method, or a `taylor1` over a hand-simplified composition ----- maximal flexibility to retain precision. I have just updated this PR to include this hook --- and we have all three methods: direct, implicit, FdB implemented as user-specifiable options.

  Repro (self-contained, plain Copulas.jl): https://gist.github.com/thisiscam/8fa9a4ea2b2dffa474cd4ea131efdc21

lrnv: Well done here. 

Thats looks great. In the documentation where you explain the problem and the possibility to overwrite the default choice, can you add a proper example while you make the argument, starting with the case of someone just trying to compute density of a NAC and obtaining 1) good results for e.g. clayton/clayton, then 2) switching to frank and obtaining bad results (show diff with bigfloats) and 3) using the hook to fix it, and 4) even proposing to implement it yourself through the third option. 

That would help clarify things for a user. 

EDIT: The faa-di-bruno version should be removed then since its performing so poor.

src/nested/NestedArchimedeanDensity.jl
```diff
@@ -0,0 +1,347 @@
...
+    gpow = copy(g)                                     # ĝ¹ (lowest order 1)
+    for k in 1:d
+        c = f[k] / T(factorial(big(k)))                # ϕ⁻¹⁽ᵏ⁾(outer,s₀)/k!
+        @inbounds for n in k:d                         # ĝᵏ has lowest order k
+            h[n] += c * gpow[n]
+        end
+        k < d && (gpow = _poly_mul_inc(gpow, g, d))    # ĝᵏ⁺¹, truncated to order d
+    end
+    return h
+end
```
lrnv: So this shoudl be removed then, thanks for trying out

src/nested/NestedArchimedeanDensity.jl
```diff
@@ -0,0 +1,347 @@
...
+    h = Vector{T}(undef, d)
+    binom = one(T)                                     # generalized binomial C(r,k), incremental
+    θ_in_pow = one(T)                                  # θ_in^k
+    for k in 1:d
+        binom    *= (r - (k - 1)) / k                  # C(r,k) = C(r,k-1)·(r-k+1)/k
+        θ_in_pow *= θ_in
+        h[k] = (θ_in_pow / θ_out) * binom * base^(r - k)
+    end
+    return h                                            # [h₁,…,h_d], length d, no constant
+end
```
lrnv: this function should move to the ClaytonGenerator file, with correct comments on top explaining what it should compute

lrnv:
@thisiscam I just wanted to tell you that i did not forgot about this, i am a bit swamped into the end of years stuff right now, but will definitely come back to you with the intention to merge. We'll talk soon :)

thisiscam:
> @thisiscam I just wanted to tell you that i did not forgot about this, i am a bit swamped into the end of years stuff right now, but will definitely come back to you with the intention to merge. We'll talk soon :)

Hi @lrnv ! No worries at all! I am also quite swamped at the moment. 
Just let me know when you get a chance to look at this. 

I hoped to get a corrected preprint of our paper that cleanly acknowledges your work and reflects the current state of Copulas.jl (which right now is a bit in flux because of this PR), and so haven't updated our arxiv yet.

lrnv:
Current CI faillure come from https://github.com/lrnv/Copulas.jl/issues/375 and are unrelated. Should be solved as soon as Roots.jl published a 3.0.4 version, whihc whouls happen quick

lrnv:
If CI goes through we are good to merge

lrnv:
@thisiscam Thansk a lot for the contribution, I will merge this now. What would you say about writing an example page in the docs (in `docs/src/examples/` i think) to showcase the censoring part of your initial work through an example ? That would IMHO be a great addition to the documentation. 

Good job :)

Issue #195, opened 2024-04-04
[bug] Dont always sample Float64's


lrnv:
I think this is an upstream problem - the `copula` (and hence also `input_distribution`) lost the type information and generate samples of type `Float64`.

_Originally posted by @devmotion in https://github.com/SciML/GlobalSensitivity.jl/pull/158#discussion_r1550752719_

PR #368, opened 2026-05-30
Fix StackOverflow in condition() for non-Float64 values

thisiscam:
`condition(C, js, uⱼₛ)` and its `SklarDist` method type the conditioning values as `NTuple{p, Float64}`. The untyped entry point routes through `_process_tuples`, which normalises with `float.` — but `float(::BigFloat) === BigFloat` and `float(::Float32) === Float32`, so non-`Float64` reals miss the typed method, fall back to the untyped entry point, and recurse forever:

```julia
julia> condition(ClaytonCopula(3, 2.0), (1, 2), (big"0.3", big"0.4"))
ERROR: StackOverflowError:
```

(also reproduces with `Float32`, and via the scalar / `SklarDist` entry points).

**Fix:** widen both typed methods to `NTuple{p, <:Real}`. `Float64` is still `<:Real`, so the existing `Float64` path is unchanged — the method is *replaced*, not added alongside, so there is no new ambiguity. Non-`Float64` values are converted to `Float64` by the downstream `DistortionFromCop`/`ConditionalCopula` fields, so conditioning is still computed in `Float64` regardless of input precision (making the conditioning machinery itself precision-generic would be a larger, separate change).

Adds a regression test covering the `Copula` and `SklarDist` entry points, scalar and tuple `js`, single- and multi-conditioned dims, with `BigFloat` and `Float32` values.

Found while building nested-Archimedean support (#367), where conditioning on `BigFloat` coordinates is natural.

### DIFF
--- a/src/Conditioning.jl
+++ b/src/Conditioning.jl
@@ -233,14 +233,20 @@ Notes
     of that conditional distribution.
 """
 condition(C::Copula{D}, j, xⱼ) where D = condition(C, _process_tuples(Val{D}(), j, xⱼ)...)
-function condition(C::Copula{D}, js::NTuple{p, Int}, uⱼₛ::NTuple{p, Float64}) where {D, p}
+# Accept any real `uⱼₛ` (not only `Float64`): `_process_tuples` calls `float.`,
+# which keeps `BigFloat`/`Float32` as-is, so a `Float64`-only signature here let
+# such inputs fall back to the untyped entry point above and recurse forever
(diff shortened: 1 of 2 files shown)


lrnv:
Thanks a lot for this. Also related is #195 : the promise of type-agnosticity on the front page of the package is not really repected in most of our code. Fixing it will require many changes and i keep pushing back ;) It's nice that NACs already support it :)

PR #369, opened 2026-06-01
subsetdims: support full coordinate permutations (p == d)

thisiscam:
`subsetdims(C, dims)` currently asserts `p < d`, so a `dims` that is a non-identity **permutation** of all `d` coordinates throws — even though it is a well-defined reordering. (This came up in the #367 review, re: the `p == d` reordering case.)

This relaxes the assert to `p <= d`. The identity `dims == 1:d` still short-circuits to `C`, and `p == 1` to a `Uniform`, so the only newly-allowed case is a genuine permutation. Invalid inputs (`p > d`, duplicate dims, out-of-range dims) are still rejected by the remaining asserts.

The generic `SubsetCopula` machinery and the per-family specializations already handle `p == d`:
- **elliptical** (`Gaussian`/`T`): reindex `Σ[dims, dims]` — respects the permutation;
- **Archimedean**: exchangeable, so any permutation yields the same copula;
- **generic**: `_cdf` saturates 0 dims (just permutes), `_logpdf` reduces to the full density, `_rand!` permutes a parent draw.

Adds `test/Subsetting.jl`: the regression (`p == d` no longer throws) plus a correctness check on an **asymmetric** Gaussian where the permutation is non-trivial (`subsetdims(C, perm).Σ == Σ[perm, perm]`, and `cdf`/`logpdf` match the permuted parent).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Subsetting.jl
+++ b/src/Subsetting.jl
@@ -81,7 +81,9 @@ This function extracts the dependence structure among the specified dimensions f
 function subsetdims(C::Copula{d},dims::NTuple{p, Int}) where {d,p}
     p==1 && return Distributions.Uniform()
     dims==ntuple(i->i, d) && return C
-    @assert p < d
+    # p == d is allowed: a `dims` that is a (non-identity) permutation reorders the
+    # coordinates. The identity `dims == 1:d` is already returned above.
+    @assert p <= d
     @assert length(unique(dims))==length(dims)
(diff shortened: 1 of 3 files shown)


lrnv:
Work for me, thanks for the catch :)

PR #371, opened 2026-06-20
add derivates for extreme value copulas

Santymax98:
This PR adds analytical implementations of dA and d²A for some bivariate extreme value tails: 

- AsymGalambosTail
- AsymLogTail
- AsymMixedTail
- MixedTail

The Tails, such as BC2Tail, MOTail and CuadrasAugeTail, continue to use ForwardDiff, since their Pickands functions are only partly differentiable.

While testing, I also found a small bug in GenericTests.jl for asymmetric extreme-value tails. The test for ℓ used ``s = y / (x + y)``, but the Tail2 interface defines ``A(tail, (u,v)) = A(tail, u)``. Therefore ``ℓ(x,y) = (x+y)A(x/(x+y))``, so the test was updated to use ``s = x / (x + y)``.

All package tests pass locally.

### DIFF
--- a/src/ExtremeValueCopula.jl
+++ b/src/ExtremeValueCopula.jl
@@ -9,7 +9,7 @@ Extreme-value copulas model tail dependence via a stable tail dependence functio
 via a Pickands dependence function ``A``. In any dimension ``d``, the copula cdf is
 
 ```math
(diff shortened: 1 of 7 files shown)


lrnv:
Please remove the filter to check only extreme values and then you are good to go I think

PR #389, opened 2026-08-18
Add Nataf correction for margin-aware Pearson correlation targets

### LINKED ISSUE #388: Nataf correction for margin-aware Pearson correlation
Hello!

I recently came across something which is referred to as the 'nataf correction' or 'nataf algorithm'. The purpose is basically to calibrate a copula parameter against a target Pearson correlation. 

From my basic understanding of copulas, is that this is nasty because the Pearson correlation parameter depends on the marginal distribution forms. The nataf algorithm is supposed to correct for this dependency.

## MWE

To demonstrate the problem here is an MWE:

```julia
using Copulas, Distributions, Statistics, Random

p = 0.7

C = GaussianCopula([1.0 p; p 1.0])
D = SklarDist(C, (LogNormal(0.0, 0.8), LogNormal(0.0, 0.8)))
x = rand(Xoshiro(1), D, 400_000)

cor(x[1, :], x[2, :])    # 0.6327
```

So even though we specified a Pearson correlation of 0.7 this is not the correlation we observe.

## Goal

It would be nice to be able to specify something like this:

```julia
D = SklarDist(GaussianCopula, margins; pearson = R)
```

## Caveats

- This formulation is only well posed for families parameterized by a matrix. For the Gaussian copula there is a fast path using Gaussian hermite, implemented by https://github.com/adknudson/PearsonCorrelationMatch.jl (*and Claude below*), but for the general case it amounts to solving a 2D integral.

- For other families parametrized by one parameter you would have to find some way for the parameter to match the summary. Or just throw an error.

I think the main question is, does such a technique belong in in the scope of Copulas.jl? 

ps. I am an electrical engineer (trying to use copulas for a package I am working) and not a statistician at all so I am not yet able to understand all the details :)

pps. I had Claude draft this implementation for the Gaussian case which I am currently using:

<details><summary>
Nataf implementation
</summary>

```julia

# The Nataf correction. A Gaussian copula with parameter rho_0 on the germ induces
# a Pearson correlation rho between the germ variables that depends on their
# marginals, and only equals rho_0 when both marginals are Gaussian. Matching a
# Pearson target therefore means inverting
#
#     rho(rho_0) = E[ g_i(Z_i) g_j(Z_j) ],   (Z_i, Z_j) ~ N(0, [1 rho_0; rho_0 1])
#
# where g_k(z) = (F_k^{-1}(Phi(z)) - mu_k) / sigma_k is the standardized marginal
# pulled back to standard normal space. The expectation is a product Gauss-Hermite
# rule, and rho is increasing in rho_0, so the inversion is a bisection.

"""
    gauss_hermite(n) -> (nodes, weights)

`n`-point Gauss-Hermite quadrature for the standard normal weight, from the
Golub-Welsch eigenvalue problem on the probabilists' Hermite recurrence. The
weights sum to one, so `sum(w .* f.(z))` approximates `E[f(Z)]` for `Z ~ N(0,1)`.
"""
function gauss_hermite(n::Integer)
    n >= 2 || throw(ArgumentError("need at least 2 quadrature nodes, got $(n)"))
    jacobi = LinearAlgebra.SymTridiagonal(zeros(n), sqrt.(1.0:(n-1)))
    factorization = LinearAlgebra.eigen(jacobi)
    return factorization.values, vec(factorization.vectors[1, :]) .^ 2
end

# The standardized marginal in standard normal space. The clamp keeps the tail
# nodes off the exact 0 and 1 that would send a quantile to infinity, which starts
# to matter beyond about 60 nodes.
function standardized_marginal(d::UnivariateDistribution, k::Integer)
    mu = mean(d)
    sd = std(d)
    (isfinite(mu) && isfinite(sd) && sd > 0) || throw(
        ArgumentError(
            "the Nataf correction needs a finite mean and a finite positive standard " *
            "deviation, but marginal $(k) ($(d)) has mean $(mu) and standard " *
            "deviation $(sd)",
        ),
    )
    return z -> (quantile(d, clamp(cdf(Normal(), z), eps(), 1 - eps())) - mu) / sd
end

# rho(rho_0) on the product rule. The conditional form z_j = rho_0 z_a + sqrt(1 -
# rho_0^2) z_b turns the correlated bivariate expectation into two independent
# standard normal integrals, so one set of nodes serves every rho_0.
function induced_correlation(gi_at_nodes, gj, z, w, rho0::Float64)
    scale = sqrt(max(0.0, 1 - rho0^2))
    total = 0.0
    @inbounds for a in eachindex(z)
        inner = 0.0
        for b in eachindex(z)
            inner += w[b] * gj(rho0 * z[a] + scale * z[b])
        end
        total += w[a] * gi_at_nodes[a] * inner
    end
    return total
end

function nataf_pair(rho::Float64, gi_at_nodes, gj, z, w, i::Integer, j::Integer)
    rho == 0 && return 0.0
    lo = induced_correlation(gi_at_nodes, gj, z, w, -1.0)
    hi = induced_correlation(gi_at_nodes, gj, z, w, 1.0)
    if !(lo <= rho <= hi)
        throw(
            ArgumentError(
                "the target Pearson correlation $(rho) for germ pair ($(i), $(j)) is " *
                "outside the range these marginals can attain, which is " *
                "[$(round(lo, digits = 4)), $(round(hi, digits = 4))]. Skewed " *
                "marginals cannot reach every correlation, so the target itself is " *
                "the thing to fix.",
            ),
        )
    end
    a = -1.0
    b = 1.0
    # rho is increasing in rho_0, so bisection cannot lose the root, and 60
    # iterations put it at machine precision on this interval
    for _ = 1:60
        b - a < 1e-12 && break
        mid = (a + b) / 2
        if induced_correlation(gi_at_nodes, gj, z, w, mid) < rho
            a = mid
        else
            b = mid
        end
    end
    return (a + b) / 2
end

"""
    pearson_to_gaussian(rho, di, dj; nodes = 32)
    pearson_to_gaussian(R, marginals; nodes = 32)

Nataf correction: the Gaussian-copula parameter that induces the target Pearson
correlation `rho` between two germ variables with marginals `di` and `dj`, or the
corrected parameter matrix for a full target matrix `R` and its `marginals`. This
is the Pearson counterpart of the closed-form [`spearman_to_gaussian`](@ref).

`marginals` is a vector of `Distributions.jl` distributions or of
[`GermVariable`](@ref)s, in germ variable order. Zero targets map to exactly zero,
and Gaussian marginals reproduce `R` exactly.

The target is on the *germ*. Since the correlation is Pearson rather than rank
based, it survives to the physical injections only through affine transforms; a
nonlinear transform changes it. Rank correlation stays the package default for
that reason.

`nodes` is the number of Gauss-Hermite nodes per dimension. The default is
accurate to about 1e-8 for the usual load and generation marginals. Heavy-tailed
or strongly skewed marginals converge more slowly and want more nodes.

Skewed marginals cannot attain every correlation. A target outside the attainable
range throws, naming the pair and the range.
"""
function pearson_to_gaussian(
    rho::Real,
    di::UnivariateDistribution,
    dj::UnivariateDistribution;
    nodes::Integer = 32,
)
    -1 <= rho <= 1 || throw(ArgumentError("correlation must lie in [-1, 1], got $(rho)"))
    z, w = gauss_hermite(nodes)
    gi = standardized_marginal(di, 1)
    gj = standardized_marginal(dj, 2)
    return nataf_pair(Float64(rho), [gi(zk) for zk in z], gj, z, w, 1, 2)
end

function pearson_to_gaussian(
    R::AbstractMatrix{<:Real},
    marginals::AbstractVector;
    nodes::Integer = 32,
)
    dists = marginal_dists(marginals)
    d = size(R, 1)
    size(R, 2) == d ||
        throw(ArgumentError("correlation matrix must be square, got size $(size(R))"))
    issymmetric(R) || throw(ArgumentError("correlation matrix must be symmetric"))
    all(isapprox.(diag(R), 1.0; atol = 1e-12)) ||
        throw(ArgumentError("correlation matrix must have a unit diagonal"))
    length(dists) == d || throw(
        ArgumentError(
            "got $(length(dists)) marginals for a $(d) by $(d) correlation matrix",
        ),
    )
    z, w = gauss_hermite(nodes)
    g = [standardized_marginal(dists[k], k) for k = 1:d]
    g_at_nodes = [[gk(zk) for zk in z] for gk in g]
    Rg = Matrix{Float64}(I, d, d)
    for i = 1:d, j = (i+1):d
        rho0 = nataf_pair(Float64(R[i, j]), g_at_nodes[i], g[j], z, w, i, j)
        Rg[i, j] = rho0
        Rg[j, i] = rho0
    end
    return Rg
end

marginal_dists(v::AbstractVector{<:UnivariateDistribution}) = v
marginal_dists(v::AbstractVector{<:GermVariable}) = [x.dist for x in v]
marginal_dists(v::AbstractVector) = throw(
    ArgumentError(
        "marginals must be univariate distributions or GermVariables, got " *
        "element type $(eltype(v))",
    ),
)

"""
    GaussianCopula(R, marginals; correlation = :pearson, nodes = 32)

Gaussian copula whose parameter is the Nataf correction of the Pearson target `R`
for the given `marginals`, in germ variable order. Passing the germ variables
themselves keeps that order right:

julia
vars = [GermVariable("a", LogNormal(0.0, 0.4)), GermVariable("b", Weibull(2.0, 8.0))]
dep = GaussianCopula([1.0 0.6; 0.6 1.0], vars; correlation = :pearson)


See [`pearson_to_gaussian`](@ref) for what the correction does, which targets are
attainable, and why the target is a property of the germ. Positive
semi-definiteness is validated after the correction, as for every other input
convention.
"""
function GaussianCopula(
    R::AbstractMatrix{<:Real},
    marginals::AbstractVector;
    correlation::Symbol = :pearson,
    nodes::Integer = 32,
)
    correlation === :pearson || throw(
        ArgumentError(
            "marginals are only used by correlation = :pearson, but got " *
            "$(repr(correlation)). Build the copula without them for :spearman " *
            "and :gaussian, which need no marginal information",
        ),
    )
    return GaussianCopula(pearson_to_gaussian(R, marginals; nodes); correlation = :gaussian)
end
```

</details>

langestefan:
Closes #388.

Implements the interface proposed in the issue discussion:

```julia
m  = (LogNormal(0, 0.8), Gamma(1, 2), Beta(1, 2))
R0 = [1 0.7 0.3; 0.7 1 0.5; 0.3 0.5 1]

R1 = Nataf(m, R0)
D  = SklarDist(GaussianCopula(R1), m)
cor(rand(D, 10^6)') # ≈ R0
```

This was implemented with help of Claude Fable.


lrnv:
For the docs, i think it should be discussed next to the gaussian copula AND next to the dependence metrics? makes the most sense for me. 

Also there's a closed form for the lognormal ? If so, can you implement the specific closed form as a method instead of only a test ? you can use despecialisation to still test the specific method against its own generic

src/Nataf.jl
```diff
@@ -0,0 +1,143 @@
...
+# by its moments *under the quadrature rule* so that comonotone margins map to
+# a correlation of exactly one on the rule itself.
+function _nataf_standardized(F::Distributions.UnivariateDistribution, k::Integer, z, w)
+    μ, σ = Distributions.mean(F), Distributions.std(F)
+    (isfinite(μ) && isfinite(σ) && σ > 0) || throw(ArgumentError(
+        "The Nataf correction is only defined for margins with a finite mean and a finite positive " *
+        "standard deviation, but margin $(k) ($(F)) has mean $(μ) and standard deviation $(σ)."))
+    # The clamp keeps extreme quadrature nodes away from the exact 0 and 1 that
+    # would send a quantile to ±∞.
+    q(t) = Distributions.quantile(F, clamp(StatsFuns.normcdf(t), floatmin(Float64), 1 - eps()/2))
```
lrnv: could you clamp with `nextfloat(T(-\infty))` and `prevfloat(T(\infty))` where `T=typeof(`some float input`)` to accomodate `BigFloat`s ? sometimes it helps

langestefan:
> Also there's a closed form for the lognormal ? If so, can you implement the specific closed form as a method instead of only a test ? you can use despecialisation to still test the specific method against its own generic

Good spot. Apparently there are 3 closed forms which we can specialize on: Normal-Normal, LogNormal – LogNormal, Normal-LogNormal. Should all be covered now. I will also update the docs

src/Nataf.jl
```diff
@@ -0,0 +1,193 @@
...
+    return clamp(ρ, -1.0, 1.0)
+end
+function _nataf_exact(Fᵢ::Distributions.LogNormal, Fⱼ::Distributions.LogNormal, ρ::Float64, i, j)
+    # r(ρ₀) = (exp(ρ₀sᵢsⱼ) - 1) / √((exp(sᵢ²) - 1)(exp(sⱼ²) - 1)), independent of the μ's.
+    sᵢ, sⱼ = Distributions.params(Fᵢ)[2], Distributions.params(Fⱼ)[2]
+    D = sqrt(expm1(sᵢ^2) * expm1(sⱼ^2))
+    _nataf_checkrange(ρ, expm1(-sᵢ * sⱼ) / D, expm1(sᵢ * sⱼ) / D, i, j)
+    return clamp(log1p(ρ * D) / (sᵢ * sⱼ), -1.0, 1.0)
+end
+function _nataf_exact(::Distributions.Normal, Fⱼ::Distributions.LogNormal, ρ::Float64, i, j)
```
lrnv: Isnt here another method missing for `::LogNormal, ::Normal` if the order of margins is reversed ?

src/Nataf.jl
```diff
@@ -0,0 +1,193 @@
...
+const _NATAF_TOL = 1e-10
+function _nataf_checkrange(ρ::Float64, lo::Float64, hi::Float64, i::Integer, j::Integer)
+    lo - _NATAF_TOL <= ρ <= hi + _NATAF_TOL || throw(ArgumentError(
+        "The target Pearson correlation $(ρ) for margins ($(i), $(j)) is outside the range " *
+        "[$(round(lo, digits=4)), $(round(hi, digits=4))] that these margins can attain. " *
+        "Pearson correlations of non-Gaussian margins cannot reach all of [-1, 1] " *
+        "(Fréchet-Hoeffding bounds), so the target itself has to change."))
+end
+
+function _nataf_pair(ρ::Float64, gᵢ_at_z, gⱼ, z, w, i::Integer, j::Integer)
```
lrnv: `_nataf_pair` and `nataf_exact` should really be several methods of the same functions, use dispatch please

src/Nataf.jl
```diff
@@ -0,0 +1,193 @@
...
+        end
+        r += w[a] * gᵢ_at_z[a] * inner
+    end
+    return r
+end
+
+# The attainable Pearson range is [ρ(-1), ρ(1)] (Fréchet-Hoeffding bounds of the
+# margins). The `tol` absorbs quadrature noise of a few ulps in bounds computed
+# numerically, so boundary targets snap to ρ₀ = ±1 instead of bisecting or throwing.
+const _NATAF_TOL = 1e-10
```
lrnv: can you use `prevfloat(one(T))` and `nextfloat(-one(T))` to bound, so that the bounds depend on the input type ?

src/Nataf.jl
```diff
@@ -0,0 +1,193 @@
...
+    end
+    R₀ = Matrix{Float64}(LinearAlgebra.I, d, d)
+    for i in 1:d, j in (i+1):d
+        ρ = Float64(R[i, j])
+        r₀ = _nataf_exact(margins[i], margins[j], ρ, i, j)
+        r₀ === nothing && (r₀ = _nataf_exact(margins[j], margins[i], ρ, i, j))
+        if r₀ === nothing
+            _quad!(i); _quad!(j)
+            r₀ = _nataf_pair(ρ, g_at_z[i], g[j], z, w, i, j)
+        end
```
lrnv: You are reinventing julia dispatch system here... make _nataf_exact and _nataf_pair be the same function ?

lrnv (dismissed):
Globally Ok, a few mistakes in the implementation. In particular, i'd like to take care of non-float64 types so that it works for BigFloats, with adapted tolerences with nextfloat() and prevfloat() boundings. Docs are OK.

src/Nataf.jl
langestefan: It was covered before, but with the dispatch method it now needs this so that was added.

src/Nataf.jl
langestefan: yeah, that's why I marked it draft, was still going over it. But you were too quick haha. Should be addressed now

lrnv:
@langestefan Good enough for me like that, if you want to take another look and decide when you want to make the PR ready. 

I basically rewrote the interface to add new fast paths, and added one or two for `Uniform` distributions, i am not aware of other closed-form existing.

langestefan:
Fixed a few more correctness bugs, nothing major. Also switched from `Roots.Bisection` to `Roots.A42`, which gives significantly better performance (anywhere between 2x-10x in my tests).

lrnv:
Thanks @langestefan this is good to go for me, i'll merge :)

PR #390, opened 2026-08-18
Add a bivariate scalar condition fast path

### LINKED ISSUE #378: Add an allocation-light bivariate condition fast path
> **LLM disclaimer:** This issue was drafted by an LLM (OpenAI Codex) after a source-code audit. The technical claims and proposed design should be reviewed and benchmarked by maintainers before implementation.

## Context

Vine and pair-copula algorithms evaluate bivariate conditional CDFs and quantiles in very tight scalar loops. The natural public formulation is:

```julia
h1(C, u, v) = cdf(condition(C, 2, v), u)
h2(C, u, v) = cdf(condition(C, 1, u), v)
hinv1(C, q, v) = quantile(condition(C, 2, v), q)
hinv2(C, q, u) = quantile(condition(C, 1, u), q)
```

For `Copula{2}`, however, the general `condition` frontend still processes tuples and computes the remaining index through the general-dimensional path. Even when the resulting `Distortion` is tiny and specialized, that setup cost matters when repeated millions of times.

This is an enabler for replacing downstream ForwardDiff + root-finding fallbacks with the canonical conditioning API.

## Proposal

Add an allocation-light bivariate specialization equivalent to:

```julia
condition(C::Copula{2}, j::Int, uj::Real) =
    DistortionFromCop(C, (j,), (float(uj),), 3 - j)
```

with normal index and support validation. Consider an internal scalar helper if that makes inlining and dispatch more reliable.

The result should continue to dispatch to all existing family-specific distortions (Gaussian, Student, Archimedean, EV, FGM, Plackett, M/W, survival wrappers, etc.).

## Acceptance criteria

- Preserve the current public semantics and numeric types.
- Avoid temporary arrays and `setdiff` on the bivariate scalar path.
- Verify `@inferred`/type stability for representative families.
- Benchmark `cdf(condition(...), x)` and `quantile(condition(...), q)` against direct `DistortionFromCop` construction.
- Add regression tests for both conditioning directions and boundary values.

## Downstream motivation

This would provide a fast universal pair-copula fallback for VineCopulas.jl while keeping family-specific mathematics centralized in Copulas.jl.

lrnv:
Closes #378.

Adds a direct condition(::Copula{2}, ::Int, ::Real) path that validates the scalar inputs and constructs the family-dispatched distortion without tuple processing or setdiff. Includes regression coverage for both directions, boundary values, numeric types, and invalid inputs.

### DIFF
--- a/src/Conditioning.jl
+++ b/src/Conditioning.jl
@@ -277,6 +277,12 @@ Notes
     The “conditional copula” is `ConditionalCopula(C, js, u_js)`, i.e., the copula
     of that conditional distribution.
 """
+function condition(C::Copula{2}, j::Int, uⱼ::Real)
+    1 ≤ j ≤ 2 || throw(ArgumentError("Conditioning index must be either 1 or 2."))
+    zero(uⱼ) ≤ uⱼ ≤ one(uⱼ) || throw(ArgumentError("Conditioning values must lie in [0, 1]."))
+    return DistortionFromCop(C, (j,), (float(uⱼ),), 3 - j)
+end
+
 condition(C::Copula{D}, j, xⱼ) where D = condition(C, _process_tuples(Val{D}(), j, xⱼ)...)
 # Accept any real `uⱼₛ` (not only `Float64`): `_process_tuples` calls `float.`,
 # which keeps `BigFloat`/`Float32` as-is, so a `Float64`-only signature here let
--- a/test/ConditionalDistribution.jl
+++ b/test/ConditionalDistribution.jl
@@ -18,6 +18,41 @@
     @test Z.m[2] == LogNormal()
 end
 
+@testset "Bivariate scalar condition fast path" begin
+    # Representative specialized and fallback distortions, including the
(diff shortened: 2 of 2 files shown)


lrnv:
Updated the regression test after reviewing the CI failure: numerical CDF equality is now checked only at interior conditioning values, while 0/1 boundaries verify construction and type preservation without invoking the unrelated MvNormalCDF boundary limitation. No benchmark was added, since the fast path is a small direct removal of the bivariate setdiff plumbing.

lrnv:
Moving this PR back to draft for investigation. The Julia CI jobs did not report an ordinary test failure; they kept running for roughly two hours and were manually cancelled. The small production fast path still looks reasonable, but the new regression tests or a code path they exercise appear capable of hanging. Before making the PR ready again, we need to isolate the exact test/call, determine whether the hang is caused by boundary evaluation, BigFloat dispatch, compilation, or recursion, and add a bounded regression that completes on both supported Julia versions.

PR #408, opened 2026-08-19
Reduce test wall time with type-aware CI sharding

lrnv:
## Goal

Reduce CI wall-clock time without removing any functional coverage from the test suite.

## Baseline

On the latest successful `main` run:

- Julia LTS tests: approximately 7m50s;
- Julia 1 tests: approximately 13m40s;
- `GenericTests.jl` alone: approximately 5m on LTS and 8m30s on Julia 1.

The per-copula timestamps show that compilation of the first instance of each concrete copula type dominates. Reducing sample counts therefore does not address the main cost.

## Experiment

- Keep ordinary local `Pkg.test()` monolithic and unchanged.
- Run specialized/core test files in one CI job per Julia version.
- Split `GenericTests.jl` into three parallel shards per Julia version.
- Assign complete concrete types, rather than individual instances, to shards so parameter variants share compilation.
- Preserve every bestiary entry and every generic test operation.

This draft exists to measure the resulting CI wall-clock time and shard balance before deciding whether the additional CI parallelism is worthwhile.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ concurrency:
 
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.group }} - ${{ github.event_name }}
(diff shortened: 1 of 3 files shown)


lrnv:
Interesting but wont be necessary for now.

Issue #333, opened 2025-10-13
[Constructors] Propose type-stable constructors


lrnv:
The constructors that we have right now are "friendly" in the sense that the dimension `d` is passed as a parameter to the copulas: 

```julia
MyCopula(d, args....; kwargs...)
```

But, sadly, since the dimension $d$ is then used as a type parameter, these constructors are not type stable. 

Without removing them, it would be nice to provide type stable ones as 
```julia
MyCopula{d}(args...; kwargs...)
```

for every model we have. Unfortunately, for Archimedeans, ExtremeValues and Archimax's, this get a bit more complicated with generators and tails... A question on discourse on this subject: https://discourse.julialang.org/t/write-a-constructor-for-these-aliases/133135/7

----------

On the other hand, our constructors are not garentied to return the right type since some of them specialize to `IndependentCopula`, `MCopula` or `WCopula` on certain parameters values. This is a behavior that we might want to remove, to encure full stability (and also simplify our fitting procedures).

lrnv:
We decide to reopen as it will simplify the tests to have constructors that are fully type stable, and the handle the specific cases in the methods later. 

We want 
```julia
ClaytonCopula{2}(0.0) isa ClaytonCopula
```

and then make the exception in each method. Its just an exception check and should not be too costly. 
```julia
function _cdf(C::ClaytonCopula, u)
    C.theta == 0 && return prod(u)
	...
end
```

On the other hand, it will remove a lot of unions from our inference and probably make compilation several times faster.

Also, we should add a regression test to check that all those wanna-be-I or wanna-be-W or wanna-be-M copulas are indeed behaving corectly, becaefore I, M and W were and otherwise it'll be breaking.

-------

The benefit is that now, `@inferred MyCopula{d}(pars...)` will always pass, but same thing for other constructs of the package: tails, generators, etc... it'll help a lot with TTFX cause the number of unions will be much smaller.

Issue #284, opened 2025-09-10
[Bug] Values of pdf's in corners of the hypercube


lrnv:
Several copulas returns NaN's on the corner of the hypercube. More genrally, as set in #276, we do not test the value of the pdf in the very corners. 

A study should be done for each copula to set correctly these values.

PR #417, opened 2026-08-21
Generalize extreme-value copulas to a multivariate STDF architecture

Santymax98:
## Context

This PR started as an audit of the extreme-value code, mainly to improve
numerical stability and address the EV-related parts of #284 and #333.

While working on this, I revisited #286 (`d > 2 EV and Archimax`), which was
closed without merging. I took the opportunity to rethink the EV architecture
before extending it to higher dimensions.

## What changed

I redesigned the extreme-value layer around the stable tail dependence
function (STDF), separating the general mathematical representation from the
specialized bivariate Pickands machinery.

This allowed me to extend several EV families to `d > 2`, including Logistic, Galambos, Hüsler-Reiss, extremal-t, Tawn, asymmetric Galambos, Mixed, Cuadras-Augé, Marshall-Olkin and discrete-spectral models, while keeping the existing optimized bivariate implementations when they are useful.

I also generalized the density and sampling infrastructure, added dimension-aware constructors, multivariate empirical EV estimation, and benchmark-based routing between available sampling algorithms.

The full test suite passes and I also updated the documentation to describe the new architecture and multivariate API.

## A small mathematical point

During this work I derived some of the multivariate extensions separately from the implementation. For most constructions I found clear references, but for a few identities I have not yet found the exact formulation in the literature.

I kept some notes/calculations aside and I think it could be interesting to review them together. Maybe everything is already known under another formulation, but perhaps there is also something worth proving or investigating more carefully.

This PR ended up being a larger expansion than I initially expected, so let me know what you think about the architecture and the mathematical choices.

If you need more details, we can discuss it 😅


Santymax98:
@lrnv The tests pass; the PR is large, and so is the change—perhaps some testing is needed, and I don't know how the tachometer works.

However, I tried to keep the architecture as close as possible; given that the bivariate approach is stable and offers improvements in certain regions, I kept it as a specific fallback.

There is one thing that bothers me a bit: the `UnivariateTail` or `UnivariateGenerator` naming convention. It isn't very statistical—it implies univariate distributions (a point some colleagues raised)—so perhaps `OneParameter...` would be more informative and consistent. Anyway, that’s just a detail to consider for the future, haha. Let me know if you think we can do anything else or if adjustments are needed.

lrnv:
@Santymax98 the Pr is huge ! At first glance: 

Regarding the `UnivariateTail` and `UnivariateGenerator`, i agree. Maybe `UniParametric` `BiParametric` etc. makes more sense. 

Regarding your implementation: When did you rebase ? Since a few days we try to get constructors as stable as possible and mainly use `MyCop{d}(params...)` as the default, and `MyCop(d, params...)` as a syntaxic sugar but not type stable. 

Also, your `Val(:bivariate)` and `Val(:multivariate)` is weird, on top of `Tail` and `Tail2` ? 

And then each model has its own sampling mechanism; there's no uniformity ? Then why not use directly the dispatch on `_rand!()` instead of adding another `_rand_ev_multivariate!()` or `_rand_ev_bivariate!()` ? All the informations you need are in the types i think. 

same idea with `_ev_logpdf_bivariate()`, why does this one even exists ?

Last idea i have (i have not checker if it applies to your code): why separate that much the bivariate and multivariate cases ? Since the dimension is known at compile time, normally the bivariate cases could just use the multivariate methods without extra latency. Unless the algorithms are really different, in which case it makes sense to subtype, but you can always do it directly in julia dispatch system: 

```julia
myoperation(::Copula{d}) where d = d
myoperation(::Copula{2}) = -2 # different behavior.
```

----------

I will take more time to do a more serious review next week, due to the size of the contribution i need to schedule my time ^^

Santymax98:
@lrnv Ah yes, there is a reason for that split 😅. While developing the multivariate part I started getting dispatch intersections between the historical Tail2 bivariate methods and the new family methods valid for arbitrary d. I kept Tail2 because it still carries the native Pickands machinery (A, derivatives, Ghoudi, etc.), even for families that are now valid in higher dimensions, so for a while I wanted the two computational paths to remain very explicit while testing their reductions against each other.

The Val(:bivariate) / Val(:multivariate) layer came from a slightly different issue: I did not want Tail2 or even d == 2 to determine the sampling algorithm. During the benchmarks, for example, Galambos, Hüsler-Reiss and extremal-t were substantially faster using their exact multivariate/spectral samplers even in dimension 2, while other families still benefited from the historical Ghoudi route. So I used the Vals as a temporary backend-selection trait while comparing and validating those algorithms.

Similarly, _ev_logpdf_bivariate() was mainly a way of keeping the historical bivariate density kernel explicitly available while checking it against the new density obtained from mixed partials of the STDF; there were a few dispatch intersections for d=2 that I resolved explicitly during that process.

But I see your point, and that’s what I expected or imagined you’d say 😊  haha: now that the algorithms are established and the specific "tail type" holds all the necessary information, the additional routing for `Val` likely offers no further advantage. Dispatching directly to `_rand!()` or `_logpdf()` should express the same options more cleanly. I can simplify that.

i will work in this... 👍

lrnv:
I understand the need for looking around and testing before deciding which algorithm is better for which cases. Yes this i exactly what i meant: now that you know which algo are more efficent for which cases, standard dispatch should be enough to route to the best one for each algorithm, each copula and each dimension. 

I will look at as soon as i have some time available, promised :)

Santymax98:
Perfect. In the meantime, I'm finishing up the proposal... take your time; I'll keep working on it.

lrnv:
# Disclaimer

To get a firtst glance at the modification of the internals, I am first reading the developer_guide.md and extremevalues.md documentations changes, so that i can understand how your API works and make comments on it. I have NOT read the rest of the code yet, but i trust what you wrote into the developper guide to correspond to your implementation :)

# From the developer_guide.md changes

## `Tail2` -> `PickandTail` ?

So, if i understood correctly, you made `Tail2` represent tails that have all the Pickands mechanisme, specific to d=2, implemented. Hence, if a model CAN do either d > 2 or d=2, it: 

1) HAS TO implement the `d>2` behaviors.
2) If `<:Tail2`, HAS TO Implement the Pickands behavior.

I propose then to rename the `Tail2` abstract type to `PickandTail` ? Will be clear that if `t::PickandTail` then i can do the bivariate pikands things with it. 

The default validity selector should then be: 

```julia
Copulas._is_valid_in_dim(::Tail, d::Int) = d >= 2
Copulas._is_valid_in_dim(::PickandTail, d::Int) = d == 2
```

Then, the methods will do something like: 

```julia
mymethod(::ExtremeValueCopula) = # the generic thing. Will concern bivariate models that are not Pickand AND multivariate models that are. 
mymethod(::ExtremeValueCopula{2, <:PickandTail}) = # the bivariate pickand thing, concerns only models that are bivariate AND which have Pickand behavior.
```

so you can clearly get the bahavior you want through dispatch only. 

## Constructor convention.

You highlight in the developper guide the `FamilyCopula(d, θ...)` but since a recent version we prefer to highlight everywhere the `FamilyCopula{d}(θ...)` constructors, which should always be the default one, since this one makes the dimension infered into the type system. 

Moreover, the implicit-dimensional constructor should be completely removed. You still highlight the existance of them in the docs, but i think that they should be removed so that people are obligated to specify the dimension when they construct a copula. I know that we have now things like `GalambosCopula(2.3)` infering by default a bivariate copula, but i'd largely prefer people using `GalambosCopula{2}(2.3)` and the syntaxic sugar `GalambosCopula(2, 2.3)` (which does not infer). If we can profit from your PR to change this once and for all for EV copulas, that'll be great. 

## Partial derivatives to get the density

### Generic version.

I have already in Conditioning.jl (lignes 11 to 31) code to get generic partial derivatives leveraging ForwardDiff. Okay, its targetted at cdfs, but it looks like it does the swame thing as yours, and unifying them would make sense. A good idea would be to create a common implementation and put it in `utils.jl`, that would make sense. But here we re-invented twice the same thing ^^

### Interface

You want an interface to provide these partials when they have a better algorithm than ForwardDiff, and you proposed `ellpartial(tail, x, I)`, `_ellpartial_signlog(tail, x, I)`, and the old `A`, `dA`, `d²A` in bivariate case. 

I think asking for both `ellpartial` and `_ellpartial_signlog` ilmplementation is too much. Ask only for one, and derive the second one from it automatically: I guess that the signlog is the best to ask for. 

Also, be sure to provide a generic implementation ! In theory, giving the `ell` function to the algorithm should be enough to get everything we need, right ? 

## Sampling 

You show in the developer guide the _rand method for matrices, which is the right one. Be sure that you do not implement the vector one, only the matrix one is needed ! 

Your comment about theneed to resolve the bivariate case for ambiguity is perfect.


## Conditionning ? Rosenblatt? 

I do not see anything about conditioning of d > 2 extreme value copulas ? I know that we have a full conditioning framework for d==2, is it expandable to higher dimensions or is it really hard to do ? 

Is the rosenblatt and inverse rosenblatt transfromations known for these copulas ? or the default version is the best we can get ? 


# From the extremevalue.md changes

## Pickand functon is more generic ? 

You say in the theory part that `By homogeneity, ``\ell`` can be restricted to the simplex to obtain the multivariate Pickands dependence function.`, and then you reduce imediately to dimension 2. Does that mean that the Pickand vision can also be done in higher dimensions ? Why restrict it to bivariate here ? 

## Small side note: 

In the `Tail2==PickandTail` cases, are you certain, absolutely certain, that the bivariate algorithms are more efficient than the multivariate ones ? If not, then maybe the multivariate version is enough on its own, and the bivariate version can simply be ditched ? Not sure, just asking. You say in the docs: 

```
Copulas.jl now uses ``\ell`` as the mathematical EV interface in arbitrary
dimension while preserving the mature bivariate Pickands machinery.
Bivariate formulas based on `A`, `dA`, `d²A`, Ghoudi sampling, conditional
distortions, and family-specific kernels remain available whenever the tail
provides them.
```

Maybe this machinery can simply be removed ? If the loss of performance is not too bad, it'll make everything much more straightforward. 

## When should i pass d ?

I alreayd said that i disagree a bit: You should *always* pass d. The default for constructor is as follow: 

```julia
struct MyCop{d,T} <: Copula{d}
    # There could be more parameters.
    par::T

    # Putting this constructor inside the struct ensures that everything will pass through it. 
    function MyCop{d}(some_inputs...)
        par = precomputations(some_inputs...) # if necessary
        @assert is_valid(par) # if necessary
        return new{d}(par)
    end
end
MyCop(d, inputs...) = MyCop{d}(inputs...) # syntaxic sugar constructor. 
MyCop(...) = ... # other constructors are possible, but all should use the inner one. 
```

Thus, the first group you highlight in the docs is OK but not the most efficient: 
```julia
# The following are OK but not type-stable:
LogCopula(d, θ)
GalambosCopula(d, θ)
MixedCopula(d, θ)
CuadrasAugeCopula(d, θ)
HuslerReissCopula(d, θ)
tEVCopula(d, ν, ρ)

# The following type-stable versions MUST exists: 
LogCopula{d}(θ)
GalambosCopula{d}(θ)
MixedCopula{d}(θ)
CuadrasAugeCopula{d}(θ)
HuslerReissCopula{d}(θ)
tEVCopula{d}(ν, ρ)
```

For the second group, its more complicated. You MUST get working the following: 

```julia
# This list MUST exists and work: 
HuslerReissCopula{d}(Γ)
tEVCopula{d}(ν, R)
TawnCopula{d}(α, weights)
AsymGalambosCopula{d}(α, weights)
BC2Copula{d}(a)
MOCopula{d}(λ)
EmpiricalEVMultivariateCopula{d}(U)

# This list should exists, will be a shame if it does not. 
HuslerReissCopula(d, Γ)
tEVCopula(d, ν, R)
TawnCopula(d, α, weights)
AsymGalambosCopula(d, α, weights)
BC2Copula(d, a)
MOCopula(d, λ)
EmpiricalEVMultivariateCopula(d, U)

# and this list you already have is OK as syntaxic sugar, but not required.
HuslerReissCopula(Γ)
tEVCopula(ν, R)
TawnCopula(α, weights)
AsymGalambosCopula(α, weights)
BC2Copula(a)
MOCopula(λ)
EmpiricalEVMultivariateCopula(U)
```


Understood ? Not sure i was very clear. Also, all of them should do the same. You highlight `The full asymmetric/subset parameterizations keep an explicit dimension:`

```julia
TawnCopula(d, dep, asy)
AsymGalambosCopula(d, dep, asy)
MOCopula(d, λ)
```

what does it mean ? Necessarily `TawnCopula(d, dep, asy)`, `TawnCopula{d}(dep, asy)` and `TawnCopula(dep, asy)` return the same thing, right ? If they do not, then its very bad interface. 


## Hussler-Reis two different parametrizations 

Accepting the diagonal only instead of the whole matrix is OK, models can have different constructors. As soon as there is a *default* one, the most generic one, that respects the behavior I just highlighted, you are good. This is already the case for GaussianCopula for example, no problem.

## The same constructor may lead to different types internally

Once again, this is OK, but be carefull not to over-use this. 

## EmpiricalEVMultivariateCopula =!= EmpiricalEVCopula

No, simply no. Use the same name, and infer dimension from the data !! 

Is the `DiscreteSpectralTail` the same idea as the `EmpiricalTail` ? Arent those two struct doing the same thing ? If yes, they should be merged into one. 

# Final comments. 

I am sure you can implement the few changes i asked here and answer the questions i had before i take a look at the code itself. The PR is huge, so we have to be methodological. Blessed be the developper guide and the docs, keep them updated with every changes you make please :D

Santymax98:
Thanks, this is very helpful. Let me answer the conceptual questions explicitly before changing the code.

**1. `Tail2` / Pickands terminology**

Yes: `Tail2` currently means “this tail has the native scalar bivariate Pickands machinery $A(t),A'(t),A''(t)$”, not simply “this family only exists in dimension 2”.

Mathematically, of course, Pickands is not intrinsically bivariate: for $d\ge 2$,
```math
\ell(x)=\left(\sum_i x_i\right)
A!\left(\frac{x}{\sum_i x_i}\right),
\qquad A:\Delta_{d-1}\to\mathbb R.
```
What is specifically bivariate is the scalar representation $A(t), \ t\in[0,1]$.

So I agree that `Tail2` is a bad name. I am leaning toward something explicit such as `BivariatePickandsTail` (or `ScalarPickandsTail`) rather than simply `PickandsTail`, precisely because multivariate Pickands functions also exist.

**2. Constructors**

Agreed. I will make

```julia
FamilyCopula{d}(params...)
```

the canonical constructor and

```julia
FamilyCopula(d, params...) = FamilyCopula{d}(params...)
```

the runtime-dimension convenience form.

I will remove the implicit “default (d=2)” behavior for scalar-parameter families. For structured parameters, dimension inference can remain only as optional sugar when it is completely unambiguous.

**3. `ellpartial`, `_der`, and the generic interface**

Agreed here too. There is unnecessary duplication with the generic mixed-derivative machinery already used in `Conditioning.jl`.

I think the clean interface should be:

```math
\boxed{\text{a new EV tail only needs to implement }\ell}
```

and generic mixed partials should be obtainable through AD.

Then families may optionally provide one numerically stable analytic hook for performance/stability. I do not think we should require both `ellpartial` and `_ellpartial_signlog`.

I will move the common mixed-partial machinery to `utils.jl` and reuse it both for EV densities and conditioning.

**4. Sampling**

Yes. The `Val(:bivariate)` / `Val(:multivariate)` layer was only useful while comparing algorithms. Now that the preferred algorithms are known, normal Julia dispatch on `_rand!` is enough.

And yes, only the matrix `_rand!` method is needed; the vector interface can be inherited from the usual `Distributions.jl` machinery.

**5. Conditioning / Rosenblatt for (d > 2)**

I checked this more carefully: we actually already have the generic machinery.

`DistortionFromCop` computes conditional distributions through mixed CDF derivatives, and the generic `rosenblatt` / `inverse_rosenblatt` implementations build sequential conditional distributions from that.

Therefore absolutely continuous multivariate EV copulas should inherit the existing framework; we do not need a new EV-specific Rosenblatt algorithm.

What I need to do is document and test this explicitly.

There is one caveat for spectral/empirical models with singular components: there a global Lebesgue density or ordinary smooth conditional derivatives need not exist, so those cases should be documented separately.

**6. Multivariate Pickands**

Yes, the current docs jump too quickly from the general STDF/simplex representation to the scalar (d=2) case.

The correct general statement is

```math
A(w)=\ell(w),\qquad
w\in\Delta_{d-1},
```
or equivalently

```math
\ell(x)=|x|_1 A\left(\frac{x}{|x|_1}\right).
```

Only for (d=2) can the simplex be parameterized by a single $t$, giving the familiar $A(t)$. I will rewrite that part of the documentation.

**7. Do we really need the bivariate machinery?**

This is the one point where I do not want to answer from intuition.

Earlier local benchmarks suggested that the answer depends on the family: for some families the historical Pickands/Ghoudi route is useful, while for Galambos/Hüsler–Reiss/extremal-t the multivariate/exact sampler can already be better even at (d=2).

But the current CI benchmarks do not directly establish that comparison.

So before keeping the specialized layer, I will benchmark in (d=2):

* scalar Pickands density vs generic STDF density;
* Ghoudi/Pickands sampling vs the family-specific multivariate sampler;
* runtime, allocations, and numerical behavior.

If the bivariate specialization gives no meaningful advantage for a family, I agree it should simply disappear.

**8. `EmpiricalEVCopula` vs `EmpiricalEVMultivariateCopula`**

Agreed for the public API: there should be only

```julia
EmpiricalEVCopula(U)
```

and the dimension should be inferred from the data.

Internally, however, I think it is useful to preserve the distinction between:

* an **estimator** built from data, and
* `DiscreteSpectralTail`, which is the resulting/general finite spectral representation.

So I will unify the public constructor without conflating those two mathematical objects internally.

---

There is also one mathematical point where I would particularly appreciate your opinion when you look at the literature.

For the multivariate extension of the historical **Mixed** model I use

```math
\ell_{\mathrm{Mixed},\theta}(x)
=(1-\theta)\sum_{i=1}^d x_i
+\theta,\ell_{\mathrm{Galambos},1}(x).
```

It is immediately a valid STDF because the class of STDFs is convex. In d=2,

```math
\ell_{\mathrm{Galambos},1}(x_1,x_2)
=x_1+x_2-\frac{x_1x_2}{x_1+x_2},
```

hence, with $t=x_1/(x_1+x_2)$,

```math
\ell_{\mathrm{Mixed},\theta}(x_1,x_2)
=(x_1+x_2)\left[1-\theta t(1-t)\right],
```

which recovers exactly the classical Mixed Pickands function

```math
A(t)=1-\theta t(1-t).
```

I found the classical bivariate Mixed model and the multivariate Galambos model in the literature, but so far I have not found this exact dimension-free extension stated in this form. I am **not** claiming novelty from that search alone, so if you recognize it under another representation I would be very interested.

**So my order for the next pass will be:**

- [x] rename/refine the Pickands capability;
- [x] normalize constructors;
- [x] unify mixed derivatives/conditioning machinery;
- [x] simplify the empirical public API;
- [x] fix the multivariate Pickands + Rosenblatt documentation;
- [x] run the explicit d=2 algorithm benchmarks and remove unnecessary specialization;
- [x] finish tests/coverage.

And yes... I will keep the developer guide and user docs synchronized with each architectural change :D

Santymax98:
Hi @lrnv,

I think I have now addressed the remaining points from your review.

For the higher-dimensional EV part, I clarified explicitly that the Pickands representation is multivariate,
```math
\ell(x)=|x|_1 A(x/|x|*1),\qquad A:\Delta_{d-1}\to\mathbb R,
```
and that only the scalar `A(t)` interface is specifically bivariate.

I also checked conditioning/Rosenblatt more carefully. The existing generic conditioning machinery already extends to smooth EV copulas in `d>2`, so I added explicit multivariate tests and documentation rather than introducing a separate EV-specific Rosenblatt implementation.

Finally, I ran the explicit `d=2` density and sampling benchmarks we discussed. The bivariate density kernels are roughly 7x–40x faster than the generic STDF partition-density route in the tested cases, and the generic route also deteriorates numerically for strong Galambos/Hüsler–Reiss dependence. On the sampling side, Ghoudi remains preferable for Logistic, while the family-specific samplers are substantially faster for Galambos, Hüsler–Reiss, extremal-t and Mixed.

The benchmarks also exposed one useful cleanup: the Logistic `d=2` density was delegating to `GumbelCopula`. I replaced that with a stable closed-form log-density; it agrees with the Gumbel and generic-STDF references to floating-point precision, uses zero allocations, and is about 14–15x faster than the generic STDF route in the representative cases.

Thanks again for the detailed review... several of these points led to real simplifications and better numerical behavior.

lrnv:
@Santymax98 thanks for the initial multivariate EV implementation. I substantially refined the branch while preserving its mathematical scope:
- Unified bivariate and multivariate representations for Hüsler–Reiss, extremal-t, AsymGalambos, Marshall–Olkin, and BC2.
- Standardized constructors around FamilyCopula{d}(...) and FamilyCopula(d, ...).
- Simplified sampling, density, fitting, and constructor dispatch; removed many routing helpers and duplicated one-use functions.
- Centralized subset-component, spectral, and numerical utilities.
- Preserved optimized bivariate paths while adding multivariate STDF, density, sampling, conditioning, and Rosenblatt support.
- Hardened numerical boundaries and fixed several CI regressions, including concrete EV fitting.
- Expanded and reorganized tests, including generic 3D EV coverage.
- Reworked the user and developer documentation and added a presentation of multivariate empirical EV estimation.


It still looks like the full conditioning/rosenblatt mechanisme is not working for all new models, but i'm working on it.

PR #434, opened 2026-08-26
Redesign the public API test architecture

### LINKED ISSUE #422: [Housekeeping] Reduce maintenance and CI costs
## Objective

Run a staged housekeeping program to minimize Copulas.jl maintenance and CI costs without reducing functional coverage or mixing unrelated behavioral changes into a single large pull request.

The guiding principle is one source of truth per mathematical or numerical mechanism. Similar-looking implementations must first be classified as accidental duplication, justified specialization, or genuinely different numerical problems.

## Working rules

- Measure before changing, especially compilation time versus warm execution time.
- Keep mechanical moves separate from numerical or API changes.
- Add characterization tests before consolidating numerical implementations.
- Preserve closed forms and justified multiple-dispatch specializations.
- Organize tests around distinct dispatch paths, not every family-operation Cartesian product.
- Use small reviewable PRs, each with an explicit maintenance or CI benefit.

## Success indicators

- fewer duplicated numerical implementations;
- clear ownership and discoverability of internal types;
- explicit mapping between representative models and dispatch paths;
- lower CI wall time and/or compilation work with equivalent functional coverage;
- removal of genuinely dead internal code and accidental API surface;
- documented follow-ups where consolidation is unsafe or premature.

The native GitHub sub-issues attached below track the independent workstreams. This issue is for coordination and aggregate progress only.

### LINKED ISSUE #424: [Housekeeping] Rationalize tests around dispatch paths
## Scope

Map the test suite to actual dispatch and numerical paths, then remove redundant family-operation combinations while preserving functional coverage.

The bestiary and generic tests should select representative models for mechanisms such as closed forms, generic Williamson transforms, discrete and continuous frailties, finite support and atoms, integer and fractional orders, conditioning, and numerical quadrature.

## Acceptance criteria

- documented test-to-dispatch-path matrix;
- distinction between unit, generic contract, and integration tests;
- deterministic stochastic regressions;
- duplicate coverage removed only after the same path is covered elsewhere;
- expensive exceptions replaced by targeted tests only when generic coverage is genuinely disproportionate;
- before/after test timings reported.

### LINKED ISSUE #425: [Housekeeping] Separate compilation costs from runtime and streamline CI
## Scope

Measure compilation cost separately from warm numerical execution and use that evidence to streamline CI. Avoid ineffective sample-size reductions when compilation dominates.

Assess per-file timing, representative first-call versus warm-call costs, test sharding, Julia-version duplication, and the number of concrete type combinations compiled by generic tests.

## Acceptance criteria

- baseline per-file and total CI timings;
- identification of compilation-bound versus runtime-bound files;
- proposed changes tied to measured bottlenecks;
- equivalent functional coverage across the CI matrix;
- before/after wall-time comparison on GitHub Actions.

### LINKED ISSUE #426: [API] Inventory the effective public and extension API
## Scope

Inventory the API that Copulas.jl exposes today before deciding what should be stabilized. This is an evidence-gathering task, not authorization for API churn.

## Inventory

- exported names and documented public-but-unexported names;
- methods inherited from or extending `Distributions.jl`;
- copula-specific operations (`condition`, `subsetdims`, Rosenblatt transforms, fitting, dependence measures, and related helpers);
- constructor forms, argument conventions, return shapes, and exceptions;
- generic fallbacks versus required family methods;
- developer extension hooks and strictly internal helpers;
- obsolete compatibility paths, unreachable methods, and accidental effective API.

For each operation, record its actual availability across representative regular, singular, conditional, high-dimensional, and fittable models. Note where availability currently depends on ad hoc test predicates or undocumented dispatch details.

## Deliverable

Produce a table classifying every relevant name or operation as:

1. stable public API candidate;
2. capability-dependent public API candidate;
3. documented extension point;
4. internal implementation detail;
5. compatibility-sensitive legacy or accidental surface;
6. dead/unreachable candidate supported by evidence.

## Acceptance criteria

- the inventory covers exports, documentation, implementation, and tests;
- current behavior is recorded without assuming the documentation is correct;
- generic fallback and specialization responsibilities are distinguished;
- compatibility-sensitive findings become focused follow-ups;
- no speculative removal or rename is bundled into the audit.

Parent: #427. Related housekeeping program: #422.

### LINKED ISSUE #428: [API] Define capability-based public contracts
## Scope

Turn the API inventory into an explicit capability-based contract. Decide which public operations are universal and which are only promised for models satisfying a mathematical or implementation capability.

## Questions to settle

- Which parts of `Distributions.jl` are guaranteed for every copula?
- Which constructors and return types are stable public behavior?
- When are `pdf`/`logpdf`, conditioning, Rosenblatt transforms, inverse transforms, fitting, and dependence measures available?
- Which public operations have a generic fallback, and which require a family implementation?
- How are capabilities represented without exposing test-only predicates as public API?
- Which current extension hooks are internal and therefore allowed to evolve?

Capabilities may depend on an instance or dimension, not only on a family type. The design must cover singular copulas and other models for which some operations are mathematically unavailable.

## Acceptance criteria

- a reviewed matrix of public operations, preconditions, results, and fallback guarantees;
- a clear distinction between public semantic contracts and internal dispatch hooks;
- a minimal internal mechanism, if needed, for querying capabilities consistently;
- consistent behavior and informative errors for unsupported operations;
- no requirement to implement operations that are mathematically undefined;
- focused follow-up issues for any breaking change rather than bundling it here.

### LINKED ISSUE #430: [API] Add capability-based public contract tests
## Scope

Build contract tests for the public API and map existing generic tests to those contracts. The goal is confidence in stable behavior without testing the full family-by-operation Cartesian product.

## Approach

- define small representative fixtures for universal, absolutely continuous, singular, conditionable, Rosenblatt-invertible, and fittable cases;
- test public entry points rather than incidental internal helpers;
- verify return types/shapes, boundaries, errors, round trips, and deterministic invariants;
- distinguish generic contract tests from family-specific numerical regressions;
- reuse the dispatch-path inventory from #424 to avoid redundant compilation-heavy coverage;
- ensure exported operations are not present only nominally but are reached through tests.

## Acceptance criteria

- a traceable public-operation-to-test matrix;
- contract coverage for each supported capability category;
- explicit tests for unsupported operations and singular cases;
- Rosenblatt/inverse and conditioning consistency tested where promised;
- fitting tested only for families and methods that declare support;
- redundant expensive combinations removed only after equivalent contract/path coverage is demonstrated;
- before/after test timing reported when the suite structure changes materially.

lrnv:
## Summary

- declare and document the SemVer-stable public API on Julia 1.11+
- replace the monolithic generic bestiary with compact public contracts and explicit mathematical cohorts
- separate constructor, fitting, component, Sklar, utility, and dispatch-path coverage
- keep family regressions active under 	est/old during their incremental migration
- report elapsed time per test group

Closes #422
Tracks #424, #425, #426, #428, and #430.


lrnv:
I think there is an opportunity to make the new test architecture more future-proof by making the package itself the source of truth for capabilities, rather than encoding those capabilities again in the test suite.

Ideally, the bestiary should only answer one question: **which representative objects do we want to exercise?** Once a copula is added to it, the test harness should infer as much as possible about the contracts it must satisfy from the type/interface structure of the package itself.

For example, if a new type is an `ArchimedeanCopula`, that fact should automatically make it go through the generic `Copula` contracts as well as the Archimedean-specific ones. Likewise for continuous copulas, elliptical copulas, `SklarDist`, generators, etc. Where a behavior is structural rather than represented by an abstract type, it may be possible to infer it from the actual method/interface implementation rather than maintaining a parallel `has_pdf`, `supports_fit`, etc. description.

The important property I would aim for is:

> Adding a new copula implementing existing interfaces should require adding representative instances/data to the bestiary, but should not require teaching the test architecture what that copula is capable of.

Otherwise, there is a risk that the various registries/cohorts gradually become a second representation of the package's architecture. The implementation says what a type can do, while the test metadata separately says what we expect it to do, and those two descriptions then have to be maintained in lockstep.

Some explicit metadata will of course remain necessary — valid constructor parameters, reference values, numerical tolerances, genuinely exceptional behavior, etc. — but ideally that metadata describes **test data and exceptions**, not the capabilities of the objects themselves.

I think a useful sanity check for the architecture would therefore be: **if we add `FooCopula` tomorrow, implementing only interfaces that already exist in the package, how many places in the test architecture need to know that `FooCopula` exists?** Ideally, apart from providing representative test instances and its genuinely family-specific tests, the answer should be close to zero.

That would make the bestiary a discovery mechanism feeding generic interface/property tests, rather than an inventory that also has to describe the expected behavior of every object.

PR #435, opened 2026-08-28
Add extensible copula hypothesis testing framework

Santymax98:
Hi Oskar,

After some time working on this, I think this is probably the last "big" PR I send for a while hahaha.

I had worked on this some time ago because of some academic needs, and I ended up adapting it to the API style we have been using in `Copulas.jl`.

The main idea is to introduce a generic hypothesis-testing framework for copulas, following a structure very similar to the fitting API.

Roughly, the design is:

```text
Hypothesis × Statistic × Calibration
                  ↓
              CopulaTest
```

So instead of having every test implemented as an isolated object, the hypothesis declares which statistics and calibration methods are available, and the implementation is selected through `Val` dispatch.

It is intentionally similar to:

```julia
_available_fitting_methods(...)
_fit(..., Val(method))
```

with the testing side using:

```julia
_available_statistics(...)
_available_calibrations(...)
_teststatistic(..., Val(statistic))
_calibrate(..., Val(calibration), ...)
```

The first available statistic/calibration is used as the default, as we already do for fitting methods.

The PR currently includes tests for:

* independence,
* exchangeability,
* radial symmetry,
* extreme-value dependence,
* goodness of fit for a specified copula,
* goodness of fit for a fitted copula model.

There are a few reusable calibration engines as well: simulation, randomization, multiplier bootstrap, and parametric bootstrap.

For composite GOF, the model is refitted inside every bootstrap sample.

The main reason I structured it this way is extensibility: adding another hypothesis or another statistic should mostly require implementing the mathematical part and declaring its capabilities, without changing the generic `CopulaTest` machinery.

I also added a manual section with the mathematical definitions of the tests and references, plus a section in the developer guide explaining how to extend the framework.

I tried to keep this PR limited to the testing API itself. Automatic copula family selection is something I worked on separately, so I will keep that in another independent PR rather than mixing both ideas here.


lrnv:
This is a huge piece of work, once again ;)

let me try to finish #434 and then i'll take a look at this. Or maybe the other way around so that i formalize this too in the tests ? not sure... 

I will definitely merge it, its a great idea but i need a bit of time :)

lrnv (commented):
I like the overall `Hypothesis × Statistic × Calibration` architecture: it fits well with the dispatch-oriented design already used elsewhere in Copulas.jl, avoids a large central dispatcher, and gives us a clean common result type.

A few changes required before merging. The most important ones are about composite GOF refitting and the treatment of ties.

## 1. Composite GOF refitting is no longer generic enough on current `main`

The current implementation refits a composite GOF model through:

```julia
function _gof_refit(M::CopulaModel, U::AbstractMatrix)
    return Distributions.fit(
        CopulaModel,
        typeof(_copula_of(M)),
        U;
        method=M.method,
        quick_fit=false,
        derived_measures=false,
        vcov=false,
    )
end
```

This assumes that `typeof(C)` contains all the information needed to reconstruct the fitted model.

That is no longer true for all copulas on current `main`.

For example, `NestedArchimedeanCopula` is now fitted from a template instance because the tree structure and the families of the inner nodes are runtime data, not information encoded in the type. Likewise, `SurvivalCopula` now stores the flip pattern at runtime rather than in the concrete type.

As a result, composite GOF can either fail to refit such models or silently refit a different model than the one that was originally estimated.

I think the refit itself should be dispatched from the fitted copula instance rather than unconditionally from its type, e.g. along the lines of:

```julia
_gof_refit(M::CopulaModel, U) =
    _gof_refit(_copula_of(M), M, U)

function _gof_refit(C::Copula, M, U)
    fit(
        CopulaModel,
        typeof(C),
        U;
        method=M.method,
        derived_measures=false,
        vcov=false,
    )
end

function _gof_refit(C::NestedArchimedeanCopula, M, U)
    fit(
        CopulaModel,
        C,
        U;
        method=M.method,
        derived_measures=false,
        vcov=false,
    )
end
```

and similarly preserve runtime configuration such as the `SurvivalCopula` flip mask.

At minimum, I would add composite GOF tests for a nested Archimedean model and a partially flipped survival copula.

More generally, if fitting methods gain additional runtime configuration in the future, GOF should reproduce the original fitting specification rather than only preserve `M.method`.

## 2. Ties are currently accepted silently

`_test_pseudos` accepts any finite matrix and then calls `pseudos(U)` unless `pseudo_values=true`.

However, `pseudos` currently uses ordinal ranks. Therefore tied observations are artificially assigned different ranks.

This matters for these tests, in particular for empirical-copula multiplier procedures. The asymptotic procedures implemented here are generally derived for continuous margins / no ties, while dedicated tie-aware bootstrap procedures are needed in the discrete or tied case.

I would not try to implement all tie-aware variants in this PR. For a first implementation, I would prefer a conservative behavior:

- detect ties in the relevant input margins;
- throw a clear `ArgumentError` for procedures that do not currently support them;
- document that continuous margins / absence of ties are required for now.

I think that is substantially safer than returning a plausible p-value without the theoretical guarantees advertised by the documentation.

## 3. Please verify the normalization of each published statistic

For GOF, the current statistic is implemented as:

```julia
s += abs2(cdf(Cn, u) - cdf(C, u))
return s / size(U, 2)
```

For the usual empirical-copula Cramér--von Mises statistic written in terms of
`√n (Cₙ - Cθ)`, integrating with respect to the empirical copula gives a sum of squared discrepancies rather than their mean.

The same question appears in the extreme-value statistic.

This does not currently change the bootstrap p-value if both the observed and bootstrap statistics are multiplied by the same constant, but it does change the value returned by `teststatistic(test)`. Since the public statistic is explicitly named `:Sn` and the documentation ties it to published definitions, I think the reported value should follow the conventional normalization.

I would check each statistic independently rather than applying a global change: for example, the normalization used by the radial-symmetry statistic may legitimately differ.

## 4. Keep extension hooks private unless we truly want to support them as public API

I like the extensibility of the implementation, but I would prefer not to make all of the internal hooks part of the public SemVer contract.

The developer guide currently presents functions such as

```julia
_available_statistics
_available_calibrations
_teststatistic
_calibrate
_simulation_sample
_randomization_sample
_multiplier_representation
_bootstrap_copula
_bootstrap_hypothesis
```

as an extension contract for contributors and downstream code.

These are all implementation-oriented `_...` functions, and I would rather keep as many of them private as possible. Making them `public` now would significantly enlarge the API surface and would constrain future changes to the testing machinery.

My preference would be:

- keep the generic public user-facing API small:
  - `CopulaHypothesis`
  - `CopulaTest`
  - the concrete public test constructors
  - `pvalue`
  - `teststatistic`
- keep most routing/calibration helpers internal;
- describe the developer-guide extension mechanism explicitly as an internal contributor API that may evolve;
- only promote an internal hook to `public` later if we find a real downstream extension use case that requires a stability guarantee.

In other words, I would preserve the extensible internal architecture without automatically turning every dispatch seam into a supported external interface.

This would also be consistent with the recent public/internal API split in Copulas.jl.

## 5. The multiplier implementation can become very memory-heavy

The multiplier representation currently materializes full `n × n` matrices, one for every permutation or every EV power.

For example, the default EV test stores three dense matrices simultaneously. At `n = 5000`, one `Float64` matrix is already about 200 MB.

The more concerning case is:

```julia
permutations = :all
```

where the number of permutations grows factorially and each one currently receives its own `n × n` matrix.

I do not think this needs a full algorithmic rewrite before the first merge, because the default `:G2` path is much more reasonable. But I would at least add a cost guard for `:all`.

Longer term, it would be preferable to stream/process one representation at a time rather than retain every matrix simultaneously.

## 6. Rebase the tests into the current test-suite structure

The PR was written against the previous test layout. Current `main` now organizes tests into API contracts, correctness tests, operations, etc.

I would integrate the new tests into something like:

```text
test/operations/hypothesis_testing.jl
```

rather than restoring another top-level monolithic test file.

The tests I would especially like to see after the rebase are:

- ties rejected/documented where unsupported;
- composite GOF for `NestedArchimedeanCopula`;
- composite GOF for a partially flipped `SurvivalCopula`;
- exact statistic normalization checks on small deterministic samples;
- RNG reproducibility for each stochastic calibration engine.

The existing CI is green on the PR branch, which is a good sign, but it predates the recent API/test-suite refactors on `main`.

## 7. Small cleanup items

A few smaller things are worth fixing during the rebase:

- `show.jl` contains unrelated CRLF/LF churn;
- `references.bib` appears to have large line-ending churn for a small number of added references;
- the new documentation still contains a `!!! info` block, whereas the current VitePress docs now preserve custom/native containers through `:::` syntax;
- in the exchangeability documentation, the displayed formula appears to add the weight to the squared discrepancy, while the implementation multiplies by the weight.

I would clean those up rather than carrying unrelated diff noise into this already large PR.

## What I would keep

A few parts look good to me and I would not change them just for the sake of changing them:

- using `StatsAPI.HypothesisTest` / `StatsAPI.pvalue` directly is appropriate; there is no need to depend on `HypothesisTests.jl`;
- the common `CopulaTest` result type is a good design choice;
- the `Val`-dispatch architecture is consistent with the rest of the package;
- centralizing the mechanics of simulation/randomization/multiplier/bootstrap calibration is a good separation of concerns;
- the Monte Carlo p-value correction used here is intentional and should not be replaced casually.

## Suggested priority

For me the blockers before merge are:

1. preserve runtime model structure/configuration in composite GOF refitting;
2. handle ties conservatively instead of silently applying continuous-margin procedures.

Then I would:

3. verify the exact normalization of the published statistics;
4. keep the internal extension hooks private and adjust the developer guide accordingly;
5. add a guard around obviously explosive multiplier configurations;
6. rebase the tests and docs onto the current repository structure.

With those changes, I think the PR would fit the package architecture well without unnecessarily expanding the stable internal API surface.

Santymax98:
Hi @lrnv , I think I’ve addressed all the points from your review.

Specifically, I updated the pull request against the current main version, fixed the composite GOF adjustment to preserve the copula structure/configuration at runtime, added conservative handling for ties, reviewed and corrected the published statistical normalizations, maintained internal extension hooks, added a safeguard for costly `permutations=:all` cases, reorganized the tests into the current suite structure, and added reproducibility checks for the random number generator.

I also cleaned up the documentation and noise in the diffs, and added comprehensive composite GOF integration tests for nested and partially inverted Archimedean survival copulas.

During the final review, I also noticed that radial symmetry randomization can induce ties in the rank grid, so I adjusted the internal reordering to use average ranks, while still rejecting tied input margins.

The focused hypothesis test suite now passes 161 out of 161 tests, and the full package test suite passes locally as well.

Let me know if anything else is needed.

lrnv:
@Santymax98 GG :) Another run by my favorite LLM gives the few following issues around composite GOF that I think should be addressed before merging.

### 1. Composite GOF with `GOFCopulaTest(M, U)` is not statistically consistent when `U` differs from the data used to fit `M`

The current constructor allows:

```julia
GOFCopulaTest(M, U)
```

and the documentation describes this as the form to use when testing a different data matrix against the fitted family.

However, the observed statistic is computed using the fitted model already stored in `M`, i.e. parameters estimated from the original fitting dataset, while every bootstrap replicate refits the model on its own bootstrap sample.

So if:

```julia
M = fit(CopulaModel, ClaytonCopula, U1)
GOFCopulaTest(M, U2)
```

the observed statistic uses `θ̂(U1)`, whereas the bootstrap statistics use `θ̂*` estimated inside each bootstrap replicate.

This is neither:

* a simple GOF test with fixed parameters, nor
* the usual composite GOF test for `U2`, where the model should first be fitted to `U2`.

I think the API should choose one clear semantic:

* either refit the same model specification on `U` before computing the observed composite statistic;
* or disallow the explicit-data form for composite models unless `U` is the fitting sample.

If the intention is to test new data against an already fixed fitted copula, that seems more naturally represented as a simple GOF test using the fitted copula itself.

### 2. Composite GOF still does not generically reproduce the original fitting specification

The new dispatch fixes the important runtime-structure issues for `NestedArchimedeanCopula` and `SurvivalCopula`, which is good.

However, the generic fallback still essentially does:

```julia
fit(
    CopulaModel,
    typeof(C),
    U;
    method=M.method,
    derived_measures=false,
    vcov=false,
)
```

This preserves the fitting method, but not necessarily the full fitting specification.

There are other fitting procedures whose behavior depends on runtime keyword arguments. For example:

* `CheckerboardCopula` may be fitted with a specific `m`;
* empirical extreme-value fits may depend on options such as `grid`, `eps`, `degree`, etc.

Those values are part of the estimator definition. If the bootstrap refit falls back to different defaults, the bootstrap is no longer reproducing the same estimation procedure used for the observed model.

I think this points to a more general abstraction being useful in the fitting layer, something along the lines of:

```julia
_refit(M::CopulaModel, U)
```

or a stored fitting specification that knows how to reproduce the original fit.

Then composite GOF could use that mechanism directly.

If introducing such a generic refit mechanism is too large for this PR, I would instead restrict composite GOF to models for which the refitting procedure is explicitly known to be reproducible, and throw a clear error for unsupported cases.

### 3. `GOFCopulaTest(M)` assumes that `M.method_details.U` already contains pseudo-observations

The shortcut currently retrieves:

```julia
M.method_details.U
```

and calls the test with:

```julia
pseudo_values=true
```

But the generic fitting API stores the input `U` it received, and some fitting methods support receiving raw data with `pseudo_values=false` and ranking internally.

That means `M.method_details.U` is not necessarily the actual pseudo-observation matrix used by the fitted estimator.

In the worst case, if the stored raw data happen to lie inside `[0, 1]`, the GOF shortcut could silently treat them as pseudo-observations and test the wrong sample.

A minimal fix would be to respect the stored `pseudo_values` metadata where available.

A cleaner long-term design would be for a fitted `CopulaModel` to explicitly retain the pseudo-observations actually used by the estimator, separately from the original input.

### 4. Small point: the exchangeability memory guard can be bypassed with an explicit permutation collection

The current cost guard is only triggered when:

```julia
permutations === :all
```

But users can also pass an explicit collection containing the same large number of permutations.

That path will still materialize one dense `n × n` matrix per permutation without hitting the guard.

I would move the memory check after `_exchangeability_permutations(...)` and base it on the actual resolved number of permutations rather than on whether the user wrote `:all`.

This is not a blocker for me, but it would make the safety mechanism more robust.

### Overall

The framework itself looks solid to me. I like the common `CopulaTest` result type, the `Hypothesis × Statistic × Calibration` decomposition, and the reuse of generic calibration engines.

The main remaining concern is composite GOF: the bootstrap should reproduce exactly the same estimation procedure used for the observed statistic. Right now that is guaranteed for the newly covered nested/survival cases, but not yet generically.

I would address points 1 and 2 before merging. Points 3 and 4 could potentially be handled together with that refit abstraction or as follow-up cleanup, depending on how the API is structured.

Santymax98:
@lrnv Thanks! I addressed the four points from the latest review:

* `GOFCopulaTest(M, U)` now refits the estimator on the tested sample before computing the observed statistic.
* `CopulaModel` now stores an internal reproducible fitting specification, so composite GOF replays the original fitting procedure and runtime configuration; unsupported custom parametrizations fail explicitly rather than silently changing the estimator.
* `GOFCopulaTest(M)` now respects the stored `pseudo_values` preprocessing metadata.
* the exchangeability memory guard now also covers explicit permutation collections.

I also added tests for these cases and updated the documentation.

lrnv:
Thanks Santiago! I've pushed a few focused commits to simplify the implementation before merging, keeping the composite GOF functionality.

- Replaced the statistic/calibration registries and `Val` routing with direct dispatch on the test procedure. Each public test currently implements one combination, so we can introduce more machinery when there is a concrete need. Unknown keywords now raise an error instead of being silently ignored.
- Removed `permutations=:all`, which could still create factorial-sized tuples despite the memory guard. Permutation and power collections are now normalized once, stored as vectors, and checked against the multiplier matrix memory limit.
- Narrowed the public API, removed the redundant GOF `kind` field, and consolidated the duplicated contributor documentation. The composite GOF and its estimator-refitting mechanism remain in place.
- Added small independent numerical checks for the multiplier processes and calibration, including a cyclic permutation to exercise the index orientation.
- Moved validation ahead of the composite GOF refit and cleaned up line endings and the bibliography.

Overall, this removes roughly 800 net lines. I also clarified the statistical variants in the documentation: in particular, the cited bivariate radial-symmetry theorem does not by itself establish validity of the implemented multivariate extension. That remains a theoretical point to resolve before claiming that guarantee.

PR #436, opened 2026-08-28
Add automatic copula family selection

Santymax98:
Hi Oskar,

This is the other independent part I mentioned in #435.

I originally developed the hypothesis-testing framework and automatic family selection together, but I think separating them makes the API much easier to review. This PR does not depend on #435, so either one can be merged first.

The idea here is to allow family selection directly through the existing fitting interface:

```julia
M = fit(CopulaModel, Copula, U; candidates=(ClaytonCopula, GumbelCopula, FrankCopula), criterion=:bic,)
```

Instead of introducing a separate model-selection object or fitting abstraction, `Copula` acts as the request to compare candidate families using the same `CopulaModel` machinery that is already available.

The supported criteria are:

* BIC,
* AIC,
* AICc,
* HQC.

The candidate fits are evaluated first without the additional inference cost, the best eligible family is selected according to the requested criterion, and the winning family is then fitted normally with the options requested by the user.

The returned object is still a regular `CopulaModel`, with the additional selection metadata stored in `method_details`.

The complete comparison can be inspected with:

```julia
selectiontable(M)
```

which keeps the candidate family, fitting method, convergence/status information, log-likelihood, number of parameters, and the information criteria for each candidate.

There are also two built-in candidate collections:

```julia
candidates=:default
candidates=:all
```

while an explicit tuple can be used when only a scientifically meaningful subset of families should be considered.

I think this is also a nice example of extending the existing fitting API without introducing another parallel interface: the result still behaves exactly like any other `CopulaModel`.

There is one particularly interesting connection with #435 that I intentionally left out of both PRs.

Once #435 and this PR are merged, a small follow-up could implement goodness-of-fit after model selection. We already had this case in the combined implementation, roughly:

```julia
M = fit(CopulaModel, Copula, U; candidates=(ClaytonCopula, GumbelCopula, FrankCopula), criterion=:bic,)

T = GOFCopulaTest(M)
```

The important point is that this is not the same as ordinary composite GOF. Under automatic model selection, each parametric-bootstrap replicate should reproduce the complete procedure:

```text
simulate from fitted model
        ↓
run family selection again
        ↓
fit the selected family
        ↓
compute the GOF statistic
```

So that follow-up would be a concrete example of the two APIs composing naturally: this PR provides the model-selection procedure, while #435 provides the hypothesis/statistic/calibration framework.

I left that integration out intentionally so both APIs can be reviewed independently first. If both are merged, I can send the selection-aware GOF part as a much smaller follow-up PR.

And no rush with this one either — especially with #434 and #435 already on your plate :)

### DIFF
--- a/docs/src/manual/fitting_interface.md
+++ b/docs/src/manual/fitting_interface.md
@@ -45,6 +45,59 @@ Returns a [`CopulaModel`](@ref) with:
 - `vcov` (if available),
 - `method_details` (a `NamedTuple` with method-specific metadata).
 
+## Automatic copula-family selection
+
+When the copula family is unknown, `CopulaModel` can select it automatically
+from a collection of candidate families:
+
+```@example fitting_interface
+Ctrue = ClaytonCopula(2, 4.0)
+Usel = rand(Ctrue, 1_000)
+
+Msel = fit(
+    CopulaModel,
+    Copula,
+    Usel;
+    candidates=(ClaytonCopula, GumbelCopula, FrankCopula),
+    criterion=:bic,
+    vcov=false,
+)
+Msel
+```
+
+The available information criteria are:
+
+- `:bic` — Bayesian information criterion,
+- `:aic` — Akaike information criterion,
+- `:aicc` — finite-sample corrected AIC,
+- `:hqc` — Hannan–Quinn criterion.
+
+With `criterion=:default`, BIC is used.
+
+Candidate fits are compared using the requested criterion, and the family with
+the smallest eligible finite value is selected. The winning family is then
+fitted once more using the inference options requested by the user.
+
+The complete comparison can be inspected with [`selectiontable`](@ref):
+
+```@example fitting_interface
+selectiontable(Msel)
+```
+
+Each row stores the candidate family, fitting status and method,
+log-likelihood, number of parameters, and all four information criteria.
+Candidates that fail to fit can be skipped with `on_error=:skip` (the default)
+or propagated immediately with `on_error=:throw`.
+
+Two built-in candidate collections are available:
+
+- `candidates=:default` uses a conservative set of commonly used parametric families;
+- `candidates=:all` considers the broader built-in parametric repertoire compatible with the data dimension.
+
+An explicit tuple of families is recommended when the scientific problem
+already restricts the plausible candidate set.
+
+
 ---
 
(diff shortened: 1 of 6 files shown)


Santymax98:
I think this PR should be revisited once #435 is resolved and merged.

Since the fitting/test infrastructure has changed on current `main`, and #435 introduces the GOF framework that will eventually interact with automatic model selection, it probably makes more sense to rebase and review this PR afterwards rather than update it twice.

I’ll back to it once #435 is settled.

lrnv:
Agreed, let's do #435 first

lrnv:
Thanks Santiago! I've rebased this onto current main and pushed three focused follow-up commits to keep the implementation small:

- Integrated the tests into the current runner under `operations/fitting_selection.jl`, preserving main's public API declarations and hypothesis-testing code.
- Removed the built-in candidate catalogues and custom selection-table type. Candidates are now explicit, and `selectiontable` returns a copy of the stored vector of comparison rows.
- Reused the winning fit instead of running its estimator a second time. The ordinary fitting path and selection now share the inference/assembly step, so requested covariance estimation still happens afterwards.
- Trimmed redundant selection metadata, preserved interrupts under `on_error=:skip`, and invalidated the single-family refitting specification on selected models. Composite GOF therefore rejects these models until a bootstrap that repeats selection is implemented.
- Added focused checks for failed/nonfinite/nonconverged candidates, interruption, the quick fitting interface, GOF rejection, and a one-call estimator check. Small test doubles cover selection decisions without expensive repeated optimizations.
- Updated the documentation to describe the explicit candidate interface and its limitations.

Static review and `git diff --check` are clean; I have not run Julia tests locally, so CI still needs to validate execution.


Following another review, I've pushed three additional focused commits:

- Preserve the winning model's convergence flag and iteration count explicitly, including when a specialized fitting method does not duplicate them in its metadata.
- Return a shallow copy from selectiontable, so sorting or modifying the returned vector cannot invalidate the stored winner index or change the model display.
- Apply on_error to candidate scoring failures as well as fitting failures, and share covariance-option validation between ordinary fitting and selection.
- Add regression checks for those cases and compare deferred covariance, coefficients, log-likelihood and diagnostics against ordinary maximum-likelihood fitting with a single Clayton candidate.

These corrections are pushed; no Julia tests were run locally. We will wait for CI before merging.

PR #437, opened 2026-09-01
Make boundary constructors type stable

### LINKED ISSUE #333: [Constructors] Propose type-stable constructors

lrnv:
Enhance type stability for boundary constructors by implementing reduced tail and generator functions. This change improves performance and reliability in handling various tail types.

Fixes #333


lrnv:
I don't think the current explosion in the number of methods is actually necessary.

<details>
<summary>Details</summary>

Some additional boundary logic is unavoidable once constructors become type-stable, but I think #437 currently puts too much of that logic too high in the stack.

The clearest symptom is `ArchimaxCopula.jl`: the PR adds separate specializations for `copula_measure_style`, `_cdf`, `_logpdf`, `_rand!`, `DistortionFromCop`, etc., then has to resolve intersections such as `IndependentGenerator × NoTail/MTail`, and BB4/BB5 add yet another layer of operation-specific delegation.

That looks like a sign that we are re-encoding the constructor-reduction graph inside each public operation.

## What is actually unavoidable

If we want

```julia
ClaytonCopula{2}(θ)::ClaytonCopula{2}
```

for every admissible `θ`, and `θ` is a runtime value, then Julia dispatch can no longer distinguish

```julia
θ = 1.5
θ = 0.0
θ = Inf
```

from the type of `C`.

So for a limit at which the internal representation genuinely degenerates, there has to be a runtime branch somewhere. There is no dispatch trick that removes that requirement without making the return type value-dependent again.

But that branch does not have to be duplicated across 8 operations × 20 families.

## I would push boundary handling down to the lowest mathematical layer possible

I think the reduction graph actually contains three very different kinds of boundaries.

### 1. Boundaries that are genuine extensions of the representation

For example:

```text
Clayton(0)   -> Independent
Gumbel(1)    -> Independent
Frank(0)     -> Independent
Galambos(0)  -> Independent
Log(1)       -> Independent
```

I would not treat those as operational reductions at all.

For instance, `ClaytonGenerator(0)` should simply satisfy the correct limiting identities:

```julia
ϕ(G, t)   == exp(-t)
ϕ⁻¹(G, u) == -log(u)
𝒲₋₁(G, d) == Gamma(d, 1)
```

and then the generic `ArchimedeanCopula` implementation should keep working.

Likewise for EV tails: if `GalambosTail(0)` represents independence, then its STDF/Pickands primitives and derivatives should have the correct extension at `θ = 0`. `ExtremeValueCopula` should ideally not even need to know that this is a boundary value.

This is especially important for the goal of the PR: `Clayton(0)` should really behave as a Clayton copula, not merely be an object that delegates all of its operations to `IndependentCopula`.

### 2. Compositional reductions

For example:

```text
BB1(θ, 1)         -> Clayton(θ)
BB6(...)          -> Joe / Gumbel
AsymLog(...)      -> Log
AsymGalambos(...) -> Galambos
BB4               -> Clayton / Galambos
BB5               -> Gumbel / Galambos
```

Here again, I would strongly avoid turning the reduction graph into

```text
constructor graph
      ↓
cdf graph
logpdf graph
rand graph
conditioning graph
...
```

which is what the PR is starting to do.

For example, BB4 now effectively does things like

```julia
iszero(δ) && return cdf(ClaytonCopula{2}(θ), u)
iszero(θ) && return cdf(GalambosCopula{2}(δ), u)
```

and then repeats the same idea in `logpdf`, `_rand!`, etc. BB5 follows the same pattern.

That feels like a code smell.

BB4 is `Archimax(ClaytonGenerator, GalambosTail)`. If the Clayton generator and Galambos tail are both correctly defined at their boundaries, then ideally the generic Archimax machinery should produce the right result by itself.

In other words:

> An identity between composite models should preferably emerge from the primitives of their components, rather than being reimplemented operation by operation.

### 3. Genuine degeneracies: `Inf -> M`, sometimes `-> W`

This is where I think a small central runtime mechanism is justified.

Examples:

```text
Gumbel(Inf)  -> M
Joe(Inf)     -> M
Clayton(Inf) -> M
Frank(Inf)   -> M
Frank(-Inf)  -> W
```

At `Inf`, there is not necessarily a clean generator/frailty/Williamson representation that can be extended in a meaningful way.

Here I would introduce a tiny concrete runtime trait, but importantly not something like the first implementation in #437 that returned

```julia
nothing | NoTail | MTail
```

and then reconstructed another copula and redispatched.

Something along these lines:

```julia
@enum LimitKind::UInt8 begin
    NO_LIMIT
    M_LIMIT
    W_LIMIT
end

@inline limit_kind(::Generator, ::Val) = NO_LIMIT

@inline function limit_kind(G::ClaytonGenerator, ::Val{d}) where {d}
    isinf(G.θ) && return M_LIMIT
    d == 2 && G.θ == -1 && return W_LIMIT
    return NO_LIMIT
end

@inline function limit_kind(G::FrankGenerator, ::Val{2})
    G.θ == Inf  && return M_LIMIT
    G.θ == -Inf && return W_LIMIT
    return NO_LIMIT
end

@inline limit_kind(G::GumbelGenerator, ::Val) =
    isinf(G.θ) ? M_LIMIT : NO_LIMIT

@inline limit_kind(G::JoeGenerator, ::Val) =
    isinf(G.θ) ? M_LIMIT : NO_LIMIT
```

The important part is that `limit_kind` always returns exactly `LimitKind`.

Not:

```julia
Union{Nothing, IndependentCopula, MCopula, ...}
```

So there is no reconstruction of another model and no dynamic redispatch over a union of unrelated types.

Then, at the Archimedean level, this can be handled once:

```julia
@inline function _cdf(C::ArchimedeanCopula{d}, u) where {d}
    kind = limit_kind(C.gen, Val(d))

    kind === M_LIMIT && return minimum(u)

    if kind === W_LIMIT
        @assert d == 2
        return max(u[1] + u[2] - 1, zero(eltype(u)))
    end

    return _archimedean_cdf(C, u)
end
```

And the same idea can be used for sampling:

```julia
function Distributions._rand!(rng, C::ArchimedeanCopula{d}, A) where {d}
    kind = limit_kind(C.gen, Val(d))

    kind === M_LIMIT &&
        return _rand_M!(rng, A)

    kind === W_LIMIT &&
        return _rand_W!(rng, A)

    return _rand_archimedean!(rng, C, A)
end
```

This gives us roughly one branch per structural Archimedean operation, instead of one specialization per family × boundary × operation.

A `UInt8`-backed enum plus a predictable branch is also very different from the earlier strategy that returned different object types and redispatched. It still needs to be benchmarked, of course, but it is much friendlier to inference.

## The same principle applies to EV families

For example:

```julia
@inline limit_kind(::Tail) = NO_LIMIT

@inline limit_kind(T::GalambosTail) =
    isinf(T.θ) ? M_LIMIT : NO_LIMIT

@inline limit_kind(T::LogTail) =
    isinf(T.θ) ? M_LIMIT : NO_LIMIT
```

But I would *not* put `Galambos(0)` into `limit_kind` if we can make `ℓ(GalambosTail(0), x)` naturally equal to `sum(x)`.

The trait should mean:

> the generic mathematical representation genuinely stops being usable here

not:

> the test graph happens to know that this boundary is equivalent to another family.

That distinction matters a lot.

## This would also clarify the role of `CONSTRUCTOR_REDUCTIONS`

The graph contains a lot of identities: simple Archimedean families, BB families, EV families, asymmetric EV families, Archimax, Liouville, Nested, and miscellaneous constructors.

But I don't think that graph should become a specification of internal dispatch.

It should remain only a correctness oracle:

> `source` and `target` must have the same semantics at this boundary.

So:

```julia
test_boundary_equivalence(source, target)
```

can tell us that `rand` fails at some boundary.

Then we inspect why.

If it is `Clayton(0)`, the fix should be something like:

> fix `𝒲₋₁(ClaytonGenerator(0), d)`.

Not:

> add `_rand!(::ClaytonCopula{d})` that detects zero and constructs an `IndependentCopula`.

If it is `Gumbel(Inf)`, then the representation really does degenerate, so an architectural `M` shortcut is justified.

I think following that discipline would remove a large amount of the method growth we currently see.

## Archimax is probably the best place to test this approach first

The current PR adds variants corresponding to things like:

```text
Archimax{Generator, NoTail}             -> Archimedean
Archimax{IndependentGenerator, T}       -> ExtremeValue
Archimax{IndependentGenerator, NoTail}  -> Independent
Archimax{..., MTail}                    -> M
```

for `measure_style`, `cdf`, `logpdf`, `rand!`, conditioning, etc., plus intersection cases.

I think there are two opportunities here.

First, `IndependentGenerator` could become a real mathematical generator rather than only a reduction marker:

```julia
ϕ(::IndependentGenerator, t) = exp(-t)
ϕ⁻¹(::IndependentGenerator, u) = -log(u)
# derivatives, Williamson transform, etc.
```

Then `Archimax(IndependentGenerator(), tail)` should ideally become an EV model through the Archimax formulas themselves, rather than because `cdf`, `rand!`, conditioning, and other operations all explicitly convert it to `ExtremeValueCopula`.

Likewise, `NoTail` is a genuine independence STDF. It should be complete enough that the generic formulas naturally reduce to the Archimedean case.

`MTail` is different: it may require a few singular paths, but those paths should live at the Archimax/EV engine level, not be repeated in BB4 and BB5.

## What I would keep from #437

I would keep:

```text
1. type-stable constructors
2. the new documented constructor contract
3. CONSTRUCTOR_REDUCTIONS
4. test_boundary_equivalence
5. mathematical extensions of Generator/Tail primitives at their boundaries
```

Then I would revisit the operation-specific shortcuts that have accumulated in the PR.

The rule I would use is:

> Handle a boundary at the lowest mathematical layer that can express the limit correctly.

And I would debug failures in this order:

```text
Generator / Tail primitive
        ↓ if impossible
Generic Archimedean / EV / Archimax engine
        ↓ if genuinely family-specific
Family operation
```

rather than:

```text
reduction graph
        ↓
special method for every public operation
```

I suspect that would let us remove a substantial fraction of the methods currently added by #437, especially the BB4/BB5 forwarding methods and a significant part of the Archimax-specific duplication.

The nice thing is that, once `test_boundary_equivalence` is reasonably complete, we can test this mechanically: remove the current shortcuts one by one and see which ones are actually required. That should give us a much cleaner way to simplify the PR than trying to predict all necessary special cases up front.


</details>

Issue #462, opened 2026-09-10
Standardize and improve public API docstrings


lrnv:
Once the public API boundary is clearly established, the docstrings attached to public symbols should themselves provide a consistent, useful description of that API.

At the moment, public docstrings vary significantly in style, precision, scope, and level of detail. Some provide substantial mathematical and behavioral context, while others are very short, implementation-oriented, inconsistent in terminology, or omit important observable behavior.

This makes it harder for users to understand what an object represents, what a function guarantees, which inputs are valid, and which edge cases or limitations they should expect.

The goal of this issue is to define a common standard for public docstrings and apply it systematically across Copulas.jl.

## Guiding principle

A public docstring should explain the **supported behavior of the public object or operation**, not merely describe its current implementation.

In particular, docstrings should prefer:

* mathematical or semantic meaning over storage representation;
* supported constructor/function forms over full concrete type signatures;
* observable behavior over internal dispatch details;
* documented guarantees over implementation anecdotes;
* public terminology and public functions over references to private helpers.

Undocumented fields, storage type parameters, internal helper types, caches, numerical backends, and contributor hooks should not appear as part of a public contract unless they are explicitly intended to be public.

## What a public docstring should answer

Depending on the object, a good public docstring should make it possible to answer most of the following questions without reading the implementation:

* What does this type or function represent?
* What are the supported ways to construct or call it?
* What do its arguments mean?
* What parameter domains or input constraints apply?
* What does it return?
* What observable behavior is guaranteed?
* How does it behave in important limiting or boundary cases?
* Which related public operations are naturally used with it?
* Are there numerical approximations or important limitations?
* What is a minimal representative example?
* Which mathematical references define the implemented object or method?

Not every docstring needs every section, but related public objects should follow a recognizable common structure.

## Proposed style

For public types and constructors, a typical docstring could follow:

```text
    PublicType(args...)
    PublicType{d}(args...)
    PublicType(d, args...)

Short description of what the object represents.

# Parameters / Arguments

Describe user-facing parameters, their meaning, and their valid domain.

# Behavior

Describe important observable semantics, including limiting cases,
dimension behavior, ordering, support, or other relevant guarantees.

# Examples

Show minimal examples using only public API.

# Notes

Document numerical limitations or other relevant caveats.

# References

Relevant mathematical references when appropriate.
```

For public functions:

```text
    function_name(args...; kwargs...)

One-sentence description of the operation.

# Arguments

Explain inputs and keyword arguments.

[...]

Issue #461, opened 2026-09-10
Audit and enforce the public API boundary in user-facing documentation


lrnv:
The package now has a more explicit distinction between its public API and its internal implementation machinery, but the user-facing documentation does not yet apply that distinction consistently.

In particular, the intended contract is primarily behavioral:

* public constructors and functions are API;
* documented mathematical semantics and observable behavior are API;
* methods that Copulas.jl adds to adopted interfaces such as `Distributions.jl`, `StatsBase.jl`, etc. may be API when documented as such;
* concrete storage representations, fields, undocumented type parameters, helper types, implementation hooks, caches, numerical backends, and internal dispatch hierarchies are not API unless explicitly documented otherwise.

This distinction matters both for SemVer guarantees and for allowing internal refactors without making the documentation accidentally promise implementation details.

## Current issues

Several parts of the documentation still blur this boundary.

For example:

* user-facing Bestiary pages sometimes include docstrings for internal or non-public abstract types;
* some public examples expose concrete type parameters such as `SklarDist{...}` even when these parameters primarily describe the current storage/dispatch representation;
* some public docstrings mention internal helpers or implementation hooks;
* Manual, Bestiary, and Example pages often mix narrative/theoretical documentation with embedded `@docs` blocks, making implementation-oriented docstrings part of the main explanatory flow;
* the same symbol may therefore simultaneously serve as a user guide, API reference, and contributor reference, although those audiences require different guarantees.

The goal of this issue is not to reduce the amount of documentation, but to make the boundary between **public behavior** and **current implementation** clear and enforceable.

## Proposed documentation structure

User-facing documentation should distinguish three roles:

1. **Manual / Bestiary / Examples**

   * narrative and task-oriented documentation;
   * mathematical explanations and practical usage;
   * examples using only supported public interfaces;
   * no dependency on undocumented storage representation.

2. **Public API reference**

   * canonical docstrings for exported/public symbols and documented adopted-interface methods;
   * precise behavioral guarantees;
   * constructor syntax and return semantics;
   * limitations and edge cases that are part of the public contract.

3. **Developer / internal documentation**

   * implementation hooks;
   * dispatch architecture;
   * internal abstract types;
   * reconstruction conventions;
   * performance-oriented details;
   * explicitly not covered by the public SemVer contract unless separately documented as public behavior.

Some duplication between narrative documentation and API docstrings is acceptable if it keeps those roles clear.

## Scope

[...]

PR #463, opened 2026-09-11
Clarify public API documentation and standardize docstrings

lrnv:
Refs #461 and #462.

This single branch handles the two documentation issues in order:

1. #461 ? audit and enforce the public/internal API boundary;
2. #462 ? standardize public docstrings after that boundary is settled.

Scope of the PR:

- removes embedded docstrings from narrative Manual, Bestiary, and Examples pages and links those pages to the canonical API references;
- standardizes and expands public docstrings, while documenting contributor-only hooks on the internal API page and in the Developer Guide;
- exposes the fitted copula or Sklar model through the public `fitteddistribution(M)` accessor instead of documenting `CopulaModel` storage fields or `method_details`;
- keeps public `See also` links within the public API and reserves internal cross-references for internal documentation;
- defines `Generator`, together with `?`, `max_monotony`, and `Distributions.params`, as a supported public extension protocol for custom Archimedean generators, while keeping derivative, inverse, radial, fitting, cache, trait, and optimization machinery internal;
- documents `SklarDist{CopulaType,Tuple{MarginTypes...}}` as the deliberate public fitting-target syntax, without exposing unrelated storage parameters or representation details;
- adds fitting, automatic model-selection, hypothesis-testing, conditioning, visualization, bestiary, and Julia/R benchmark documentation;
- extends the PartitionedDistributions integration so that compatible vector-valued distributions provide `Copulas.subsetdims`, `Copulas.condition`, `Copulas.rosenblatt`, and `Copulas.inverse_rosenblatt` through their public marginal/conditional interface;
- tests the extension's Rosenblatt and inverse Rosenblatt behavior with both multivariate Gaussian and multivariate Student distributions, including vector and matrix round trips and dimension validation;
- makes the docs build reject future `@docs`/`@autodocs` blocks in narrative sections.

The pre-existing FGM endpoint implementation inconsistency is intentionally not changed here; this PR only corrects the corresponding documentation.

The full documentation build is intentionally left to CI because it is expensive locally.


lrnv:
Please update PR #463 with the following decisions and corrections.

1. **Fix the incorrect FGM documentation, but do not fix the implementation bug in this PR.**

The new `FGMCopula` documentation currently states that, in dimension two, `θ = ±1` reaches the Fréchet bounds. This is mathematically incorrect for the FGM family.

For the bivariate FGM copula,

$$
C_\theta(u,v)=uv\left[1+\theta(1-u)(1-v)\right],
$$

the admissible endpoints `θ = ±1` are still ordinary FGM copulas and do not equal the upper or lower Fréchet bounds.

Remove every documentation statement claiming that `θ = 1` gives `MCopula` or that `θ = -1` gives `WCopula`, including wording such as “reaches singular Fréchet bounds”.

There appears to be a pre-existing implementation inconsistency in `FGMCopula.jl`, where `_cdf` special-cases `θ = ±1` as the Fréchet bounds. Do **not** fix that behavioral bug in this documentation PR. Keep the scope limited to correcting the documentation; the implementation issue will be handled separately.

2. **Audit public docstrings so that their `See also` sections do not expose internal API.**

Public docstrings must only point users toward other public API unless there is a very strong reason otherwise.

In particular, public docstrings such as `GaussianCopula` and `TCopula` currently link to `EllipticalCopula`, even though `EllipticalCopula` is explicitly an internal implementation type.

Perform a systematic audit of the `See also` sections of all public docstrings changed by this PR:

* remove links to internal abstract types, private helpers, internal capability traits, implementation hooks, and non-public machinery;
* replace them with the closest meaningful public concepts or concrete public families;
* keep internal-to-internal cross references in internal docstrings and the internal reference page.

The intent is that a user reading a public docstring should not be led into implementation-only API.

3. **Treat `Generator` as a real public extension protocol and revise the documentation consistently.**

Decision: the Archimedean generator interface is a supported public extension protocol.

Users are intentionally allowed to define custom generators by subtyping `Generator` and implementing the documented mathematical interface. Therefore, review all documentation around `Generator` and make it internally consistent with that decision.

In particular:

* `Generator` itself is public and supported as an extension point;
* the public extension protocol should clearly document which operations a downstream implementation is expected to provide;
* `ϕ`, `max_monotony`, and `Distributions.params` should be documented as the core supported requirements where appropriate;
* clearly distinguish this supported mathematical extension interface from internal machinery such as derivative caches, fitting hooks, radial implementation details, dispatch traits, specialized derivative methods, and other performance internals;
* the `ArchimedeanCopula` custom-generator example should remain public and should be improved if needed rather than moved to the Developer Guide;
* the Developer Guide may explain implementation details and optional optimization hooks, but it must not imply that the basic custom-`Generator` extension mechanism is unstable or contributor-only;
* review wording in the `Generator` docstring and related pages that currently says or implies that the whole subtype hierarchy is internal. Narrow that statement so it applies only to genuinely internal subtypes/hooks, not to the public `Generator` extension contract.

The result should make the boundary explicit: implementing a mathematical generator is supported public API; relying on Copulas.jl's internal helper hierarchy or optimization machinery is not.

4. **Explicitly document `SklarDist{CopulaType,Tuple{MarginTypes...}}` as a deliberate public exception.**

Decision: this type-parameter syntax is part of the supported fitting interface.

The documentation generally says that concrete storage type parameters are not public API. Keep that principle, but explicitly document the following exception:

`SklarDist{CopulaType,Tuple{MarginTypes...}}`

is a supported public way to specify the copula family and marginal families when fitting a Sklar model.

Make the distinction precise:

* the copula-family and margin-family parameters used in this fitting-target syntax are public;
* this does **not** make arbitrary implementation/storage type parameters of `SklarDist` public;
* users should not rely on additional concrete representation details beyond the documented fitting-target form;
* mention this exception wherever the general “type parameters are implementation details” rule would otherwise appear contradictory.

5. **Keep the new PartitionedDistributions Rosenblatt support in this PR, add it explicitly to scope, and strengthen its tests.**

Decision: do not split this feature into another PR.

Update the PR description/scope so that it explicitly mentions that the PartitionedDistributions extension now provides:

* `Copulas.rosenblatt`
* `Copulas.inverse_rosenblatt`

for compatible vector-valued distributions supported through the marginal/conditional interface.

Keep the existing multivariate Gaussian tests, but add at least one additional independent distribution-family test so the extension is not validated only against `MvNormal`.

Prefer a multivariate Student distribution if it is supported cleanly by the current PartitionedDistributions API. The additional test should exercise the public behavior rather than implementation internals and should cover, where applicable:

* forward Rosenblatt transformation;
* inverse transformation / round-trip;
* vector input;
* matrix input;
* dimension validation.

Avoid duplicating excessive test logic if a compact second-family test is sufficient.

6. **Do not document in-place mutation of Gaussian/Student correlation inputs as public behavior.**

The new `GaussianCopula` and `TCopula` docstrings currently explain that the supplied matrix is normalized to correlation scale in place and advise users to pass `copy(Σ)` if they want to preserve it.

Remove this implementation detail from the public docstrings.

The public documentation should describe the accepted mathematical input and validation semantics, e.g. that covariance-like inputs are normalized to a correlation matrix where applicable and that invalid/non-positive-definite inputs are rejected, without promising that mutation of the caller's matrix is part of the API.

Do not change constructor behavior in this PR unless required for correctness. The goal here is only to avoid turning the current mutation behavior into a SemVer-level public guarantee.

Please keep the other current choices in PR #463 unchanged unless one of the changes above requires a small consistency edit.

lrnv:
Please address the following three remaining review points in PR #463.

These are not general documentation-cleanup requests: each point concerns a specific public-API contract that the PR is now making explicit. Please preserve the current overall direction of the PR and make the smallest coherent changes needed to resolve them.

## 1. `CopulaModel` still exposes concrete storage fields as if they were public API

### Problem

One of the central principles established by #461 is that the public API should be behavioral rather than tied to concrete object representation.

However, the new `CopulaModel` documentation still explicitly presents its struct fields as public “Result properties”, including names such as:

* `result`
* `n`
* `ll`
* `method`
* `vcov`
* `converged`
* `iterations`
* `elapsed_sec`
* `method_details`

The narrative fitting documentation also still directly accesses some of these fields, for example:

```julia
M.ll
Ŝ.result
```

and discusses `method_details` as something users may inspect.

This conflicts with the API-boundary work being done elsewhere in the PR. If these fields remain documented this way, we effectively turn the current `CopulaModel` storage layout into a SemVer-level compatibility promise.

That is especially problematic for `method_details`, whose structure is estimator-specific and already contains internal information such as fitting specifications, covariance metadata, retained fitting data, optimizer information, etc.

### Required direction

Review `CopulaModel` from the same behavioral-API perspective used for the rest of the package.

Public documentation should prefer stable accessors and model-interface methods over concrete fields whenever an accessor already exists.

For example, document and use the appropriate existing public interfaces such as:

```julia
StatsBase.nobs(M)
StatsBase.coef(M)
StatsBase.coefnames(M)
StatsBase.vcov(M)
StatsBase.stderror(M)
StatsBase.confint(M)
StatsBase.deviance(M)
StatsBase.nulldeviance(M)
StatsBase.nullloglikelihood(M)
StatsBase.aic(M)
StatsBase.bic(M)
StatsBase.predict(M; ...)
StatsBase.residuals(M; ...)
selectiontable(M)
```

Do not document `method_details` as a public user-facing interface. Its contents should remain implementation/estimator metadata unless a particular piece of information is exposed through a dedicated public accessor.

### Fitted distribution access

The fitted distribution itself obviously needs a supported way to be retrieved.

Please make an explicit API decision here rather than accidentally stabilizing the entire struct layout.

Either:

1. explicitly declare and document `M.result` as a deliberately supported public property;

or, preferably if appropriate for the package design,

2. expose the fitted distribution through a dedicated public accessor and use that accessor throughout user-facing documentation.

The important point is that the supported behavior must be intentional. Merely having a field with a convenient name should not automatically make every `CopulaModel` field public.

The same reasoning applies to `method`, convergence state, timing information, etc. If these are intended to be user-facing guarantees, expose/document them intentionally. Otherwise keep them as display/internal metadata.

### Concrete changes to make

Please audit:

* the `CopulaModel` docstring;
* the fitting manual;
* examples using fitted models;
* related `See also` sections;
* any other new documentation added in this PR that accesses `CopulaModel` fields directly.

Replace direct storage-field discussion with public model-interface behavior wherever possible.

For example, avoid a section like:

```text
Result properties:
- result
- n
- ll
- method
- vcov
- method_details
...
```

and instead explain what information the model exposes through its public statistical-model interface.

It is fine to mention conceptually that a fitted model retains the fitted distribution, observation count, likelihood, uncertainty information, convergence diagnostics, etc.; the issue is tying those guarantees to the current field layout.

### Acceptance criteria

After the change:

* `CopulaModel`'s concrete field layout is not implicitly declared public.
* `method_details` is clearly internal/estimator metadata, not a public extension surface.
* Existing StatsBase accessors are used in user-facing examples where applicable.
* Retrieval of the fitted distribution has one explicitly documented supported API.
* Narrative docs no longer use raw fields merely because they currently exist.
* The resulting documentation remains practical: users can still clearly learn how to inspect and use a fitted model without knowing how it is stored.

---

## 2. The reverse `Copulas.condition` → `PartitionedDistributions.conditional` adapter violates the documented PartitionedDistributions contract for some distributions

### Current implementation

The extension currently implements a broad fallback roughly of the form:

```julia
function condition(
    dist::Distributions.Distribution{
        Distributions.ArrayLikeVariate{1}
    },
    js,
    xjs,
)
    ...
    x = fill(zero(T), d)

    for ...
        x[conditioned[k]] = observed[k]
    end

    return PartitionedDistributions.conditional(
        dist,
        x,
        keep,
    )
end
```

The rationale in the extension is that Copulas.jl only receives values for the observed coordinates, whereas the public `PartitionedDistributions.conditional` API requires a complete point. The coordinates being retained are therefore filled with placeholder zeros.

Mathematically, the resulting conditional law should depend only on the observed coordinates, and the current `MvNormal` and multivariate Student implementations behave that way.

However, this is stronger than the public PartitionedDistributions contract allows us to assume.

### Contract mismatch

The public PartitionedDistributions documentation states that for

```julia
conditional(dist, x, keep_indices...)
```

the full point `x` must be in the support of `dist`.

Therefore, constructing an arbitrary complete point by filling unobserved/kept coordinates with zero is not generically valid.

Examples where this can fail include distributions with support such as:

* strictly positive coordinates;
* bounded coordinates not containing zero;
* transformed multivariate distributions;
* distributions with coordinate-dependent support;
* potentially mixture/product distributions where the retained coordinates impose support constraints.

Even if the implementation of a particular `conditional` method currently ignores the retained coordinates, calling the public API with an out-of-support point violates its documented precondition.

This matters even more now because the new generic `rosenblatt` and `inverse_rosenblatt` methods call this adapter repeatedly. The PR description now promises support for “compatible vector-valued distributions”, while the generic method currently dispatches on the very broad:

```julia
Distribution{ArrayLikeVariate{1}}
```

### Why replacing zero with another arbitrary placeholder is not enough

Please do not simply replace zeros with means, medians, ones, or another heuristic value.

There is no generic public guarantee that any such value belongs to the support, and for an arbitrary vector-valued distribution we may not be able to construct a valid full support point from the Copulas.jl `condition(dist, js, xjs)` arguments alone.

The semantic mismatch is fundamental:

```julia
Copulas.condition(dist, observed_indices, observed_values)
```

contains only the information mathematically required for conditioning.

By contrast:

```julia
PartitionedDistributions.conditional(dist, full_point, keep_indices...)
```

currently requires a complete support point.

### Required direction

Please resolve this at the API-contract level rather than relying on implementation behavior that is not guaranteed by PartitionedDistributions.

There are several acceptable solutions, depending on what the available public PartitionedDistributions interface permits.

#### Preferred solution

If PartitionedDistributions exposes, or can expose, a public API that accepts only the actually conditioned coordinates/values, use that API for the reverse adapter.

That would give a clean semantic mapping:

```text
Copulas condition:
    observed indices + observed values

        ↓

PartitionedDistributions partial-conditioning API:
    observed indices + observed values
```

without fabricating missing values.

#### Otherwise: restrict the fallback

If no such public interface currently exists, narrow the generic reverse adapter to distributions for which the required full support point can be constructed with a documented guarantee.

The current broad method on all vector-valued `Distribution`s should not claim more than can be guaranteed.

It is acceptable to support explicitly known compatible families such as the current multivariate Gaussian and Student cases if necessary.

The same restriction must then propagate naturally to the generic Rosenblatt methods that depend on this adapter.

#### Another acceptable upstream resolution

If the intended semantics of PartitionedDistributions are actually that values at retained coordinates are semantically ignored and therefore need not satisfy support constraints, that should be made explicit in the PartitionedDistributions public contract first.

The Copulas extension should not silently assume a weaker precondition than the dependency documents.

### Tests

The current Gaussian and Student tests are useful and should remain.

However, they only exercise distributions with full real support, so they cannot reveal this contract violation.

Depending on the chosen implementation:

* if the adapter is narrowed, add tests demonstrating the intended supported dispatch boundary;
* if a proper public partial-conditioning API is used, add at least one test with a distribution whose support does not contain zero, so the regression specifically proves that no invalid placeholder point is fabricated;
* ensure both `condition` and the Rosenblatt round trip use the corrected path.

### Documentation

Update the PartitionedDistributions interoperability documentation so that it describes the actual supported scope accurately.

Avoid wording that suggests every vector-valued `Distributions.jl` distribution becomes compatible merely by loading PartitionedDistributions unless that is genuinely guaranteed.

The documentation should explain the capability in terms of the public marginal/conditional interface actually required.

### Acceptance criteria

After the change:

* no call to `PartitionedDistributions.conditional` relies on an arbitrarily fabricated out-of-support complete point;
* the extension's method signatures do not promise compatibility more broadly than the public dependency contract permits;
* Rosenblatt support has the same clearly defined compatibility boundary;
* Gaussian and Student tests continue to pass;
* the chosen solution is based only on public PartitionedDistributions API, not private `_conditional_impl` methods;
* documentation accurately reflects the supported scope.

---

## 3. Add a minimal contract test for the newly public `Generator` extension protocol

### Problem

The PR now deliberately establishes the following as a supported public extension API:

```julia
abstract type Generator end

ϕ(G, t)
max_monotony(G)
Distributions.params(G)
```

The documentation consistently states that a downstream package/user may subtype `Generator` and implement exactly these three mathematical operations.

It also correctly distinguishes them from internal optional machinery such as:

* `ϕ⁻¹`
* derivative hooks;
* inverse-derivative hooks;
* higher derivatives;
* inverse Williamson implementation details;
* radial caches;
* fitting hooks;
* dispatch traits;
* performance specializations.

This is a good API decision, but it should now be protected by an explicit test.

The existing custom/test generator coverage is not sufficient for this purpose because the test generator currently also implements several of the internal optional hooks, including inverse and derivative methods. Therefore those tests do not prove that the documented three-method public contract is actually enough for the behavior we promise.

### Required test

Add a deliberately minimal generator type in the test suite that implements **only** the documented public protocol.

For example, something along the lines of:

```julia
struct MinimalPublicGenerator <: Copulas.Generator end

Copulas.ϕ(::MinimalPublicGenerator, t) = exp(-t)

Copulas.max_monotony(::MinimalPublicGenerator) = Inf

Distributions.params(::MinimalPublicGenerator) = (;)
```

Do not implement any of:

```julia
ϕ⁻¹
ϕ⁽¹⁾
ϕ⁽ᵏ⁾
ϕ⁽ᵏ⁾⁻¹
𝒲₋₁
frailty
composition_taylor
fitting hooks
internal traits
```

The point of this test is specifically to ensure that no internal method is accidentally required by the basic public extension contract.

### What to test

At minimum:

```julia
G = MinimalPublicGenerator()
C = ArchimedeanCopula(3, G)
```

must construct successfully.

Then verify the generic CDF path. Since `ϕ(t) = exp(-t)` corresponds to independence,

```julia
u = [0.2, 0.5, 0.8]
@test cdf(C, u) ≈ prod(u)
```

is a strong independent oracle.

Also test the public parameter interface:

```julia
@test params(G) == (;)
```

and, if useful:

```julia
@test length(C) == 3
```

The main requirement is construction + CDF using no private generator hooks.

### Optional additional test

If the documentation deliberately promises that another operation works generically from the same minimal contract, test it as well.

However, do **not** expand the public promise merely to make tests more ambitious.

The current wording carefully says that the three methods guarantee construction and the generic CDF, while density/sampling/etc. may additionally depend on regularity and internal numerical fallbacks. Preserve that distinction.

It is perfectly acceptable for this minimal contract test to stop at CDF.

### Location

Prefer placing the test near the Archimedean/generator correctness tests rather than embedding it in an unrelated nested-copula test.

A small dedicated testset such as:

```julia
@testset "public Generator extension contract" begin
    ...
end
```

would make the intent very clear.

### Why this matters

This PR is turning the generator extension mechanism into a real public SemVer contract.

The test should therefore operate like a compatibility test for downstream users:

> If a downstream package defines a subtype of `Generator` and implements only the three methods that Copulas.jl documents as required, future internal refactors must not accidentally make a private hook mandatory for basic construction/CDF behavior.

That is exactly the kind of invariant #461/#462 are trying to make explicit.

### Acceptance criteria

* A custom `Generator` subtype implementing only `ϕ`, `max_monotony`, and `Distributions.params` is covered by tests.
* `ArchimedeanCopula(d, G)` constructs successfully from that generator.
* Generic CDF evaluation is tested against an independent known result.
* The test does not implement any optional/internal generator hooks.
* No additional unsupported capability is accidentally promoted to the public contract.
* The test clearly communicates that it protects the downstream public extension interface.

PR #464, opened 2026-09-12
Release version 1.0.0

lrnv:
# Release Copulas.jl 1.0.0

This PR bumps Copulas.jl to `1.0.0`.

The package now has an explicitly documented public API, with a clear boundary
between stable user-facing behavior and implementation details. Julia 1.11 is
the new minimum supported version. The PR remains a draft while we wait for a
final review from @Santymax98.

<details>
<summary>Draft release notes</summary>

# Copulas.jl v1.0.0

[Documentation](https://lrnv.github.io/Copulas.jl/stable/) ·
[Repository](https://github.com/lrnv/Copulas.jl) ·
[Full diff since v0.1.42](https://github.com/lrnv/Copulas.jl/compare/v0.1.42...v1.0.0)

Copulas.jl 1.0 is the first release with an explicitly defined, documented,
and SemVer-stable public API. Not every capability below was introduced by the
version-bump commit: this release consolidates the substantial evolution of the
package throughout the `0.1.x` series.

## Compatibility contract

The documentation now separates the stable public API from explanatory manuals
and implementation-oriented developer documentation. Public compatibility is
defined by exported or `public` symbols and their documented behavior.

Concrete fields, caches, numerical backends, internal traits, optimizer
coordinates, and underscored fitting or conditioning machinery remain
implementation details. In particular:

- fitted distributions are retrieved from `CopulaModel` with
  `fitteddistribution`;
- fitting is exposed through `fit`, with `fitting_methods` reporting the
  estimators supported by a family;
- fitting internals are not a downstream extension protocol;
- `Generator`, `ϕ`, `max_monotony`, and `Distributions.params` form the public
  protocol for custom Archimedean generators;
- `SklarDist{CopulaType,Tuple{MarginTypes...}}` is supported as a fitting target
  without exposing the remaining storage representation.

Copulas.jl 1.0 requires Julia 1.11 or later. Users requiring older Julia
versions should remain on a compatible `0.1.x` release.

## Fitting, inference, and model selection

The fitting API now distinguishes maximum likelihood (`:mle`) from maximum
pseudo-likelihood (`:mpl`) and gives pseudo-observations an explicit,
order-invariant tie policy with configurable ranking methods.

`CopulaModel` is deliberately minimal. It retains the fitted distribution, the
original fitting data, the fitted log-likelihood, and the recipe required for
reproducible refitting. Coefficients, parameter names and blocks, transformed
copula observations, and the independence likelihood are derived when needed.
Optimizer traces, convergence records, elapsed times, objectives, transformed
copies of the sample, and covariance information are not persistent model
state.

Uncertainty is computed after fitting with `infer`, which returns a separate
`CopulaInference` supporting covariance matrices, standard errors, and
confidence intervals where the requested procedure is defined. Bootstrap and
jackknife procedures replay the recorded estimator, including complete
sequential Sklar fits.

Automatic family selection compares an explicit candidate set using AIC, BIC,
AICc, or HQC. It returns a `CopulaSelection`; `selectedmodel` retrieves the
winning ordinary `CopulaModel`, while `selectiontable` reports successful,
failed, and non-finite candidate fits. Selection history is not inserted into
the winning model.

The representation leaves room for future fixed-parameter fitting without
storing duplicate parameter trees. Full generic Sklar maximum likelihood is
not claimed: marginal `fit` implementations do not expose a common likelihood
or unconstrained-parameterization contract.

## Statistical semantics corrected before 1.0

The pre-release audit resolved several boundary and interpretation issues:

- Bernstein constructions now enforce the conditions required for exactly
  uniform margins and no longer truncate valid densities above one;
- the finite-sample `EmpiricalCopula` is documented and handled according to
  its atomic empirical-distribution semantics;
- nested Archimedean constructors validate supported nesting conditions;
- Gaussian copula MLE maximizes the copula likelihood;
- bivariate FGM copulas at `θ = ±1` remain absolutely continuous FGM models
  rather than being replaced by Fréchet bounds;
- survival and partial-reflection semantics now include explicit bivariate
  rotation types and public `basecopula`, `flipmask`, and `flips` accessors.

## Copula families and mathematical components

The package exposes a broad bestiary through a common `Distributions.jl`
interface:

- Archimedean families including Clayton, Frank, Gumbel, Joe,
  Ali–Mikhail–Haq, inverse Gaussian, BB1–BB10, and custom generators;
- multivariate Gaussian and Student copulas;
- nested Archimedean and Liouville copulas;
- logistic, Galambos, Hüsler–Reiss, extremal-t, Tawn, asymmetric,
  Marshall–Olkin, and discrete-spectral extreme-value models;
- empirical, beta, Bernstein, checkerboard, and empirical extreme-value models;
- bivariate Archimax constructions and named BB4/BB5 forms;
- independence and Fréchet bounds, reflections, rotations, subsets, and other
  compositional models.

The extreme-value layer is organized around stable tail dependence functions,
with specialized bivariate Pickands representations where useful. Liouville
copulas support arbitrary positive Dirichlet parameters, and Student copulas
include multivariate CDFs, conditioning, rank matching, and non-integer degrees
of freedom.

## Common operations

Copulas and Sklar distributions participate in common high-level operations:

- CDF, density, likelihood, and random sampling;
- `subsetdims` and `condition`;
- `rosenblatt` and `inverse_rosenblatt`;
- hyperrectangle probabilities through `measure`;
- Kendall, Spearman, Blomqvist, Gini, entropy, and tail-dependence measures;
- resampling-based tests for independence, exchangeability, radial symmetry,
  extreme-value dependence, and simple or composite goodness of fit.

Conditionals are represented as distributions. Their public interface is
`condition` followed by operations such as `cdf`, `quantile`, `pdf`, or `rand`;
family-specific h-functions and derivative kernels remain internal.

## Ecosystem integration

Package extensions provide plotting recipes, ExpectationMaximization.jl
adapters, and interoperability with PartitionedDistributions.jl. Compatible
non-copula random vectors can reuse subsetting, conditioning, Rosenblatt, and
inverse Rosenblatt operations through that extension. `Nataf` provides latent
Gaussian correlation correction for prescribed non-Gaussian margins.

## Documentation, testing, and performance

The Manual explains workflows, the Bestiary documents definitions and model
caveats, Examples cover applied tasks, the Public API is the canonical stable
reference, and the Developer Guide documents internals without granting them a
SemVer guarantee.

The test suite is organized around public behavioral contracts, independent
mathematical oracles, implementation-branch coverage, extension tests, and
targeted regressions. Tachometer tracks representative performance paths, while
reproducible Julia/R benchmark code is included in the documentation.

## Migrating from 0.1.x

- Upgrade to Julia 1.11 or later.
- Use documented constructors and semantic accessors instead of concrete fields.
- Retrieve fitted distributions with `fitteddistribution`.
- Obtain uncertainty from `infer(model)` and the resulting `CopulaInference`.
- Retrieve automatic-selection winners with `selectedmodel`.
- Use `fit` and `fitting_methods`; do not depend on underscored fitting hooks.
- Use `condition` with distribution operations instead of internal h-functions.
- Use `params`, `basecopula`, `flipmask`, and `flips` instead of storage layout.

Version 1.0 is a stability milestone rather than an end point. New families,
inference methods, performance specializations, and additive functionality can
continue to arrive throughout 1.x while preserving these contracts.

Thank you to everyone who contributed code, mathematical review, bug reports,
documentation, benchmarks, and API discussion on the road to 1.0.

</details>

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Copulas"
 uuid = "ae264745-0b69-425e-9d9d-cf662c5eec93"
-version = "0.1.42"
+version = "1.0.0"
 authors = ["Oskar Laverny"]
 
 [deps]


Santymax98:
Hi @lrnv  excellent initiative. I’m fully on board.

Just let me look into a few minor bugs and a couple of pending items I’d noted as potential issues—things I’d like to review before the 1.0 release (nothing too complex at this stage). I’d like to use this as a springboard for ideas regarding possible JSS extensions and, perhaps, even a future book project together.

I’d also like to compare the current state of `Copulas.jl` with that of `VineCopulas.jl` to ensure it doesn't fall behind.

Give me until the end of the day (Recife time), and I’ll send you my ideas.

Santymax98:
@lrnv I examined the statistical side in somewhat greater detail before version 1.0.

I still think releasing 1.0 is a very good idea. I do not think we should wait for new features, and I agree that Copulas.jl stopped being merely a collection of copula families a long time ago and has become a much broader modeling framework.

However, I think there are three different kinds of things worth separating and discussing before proceeding.

# 1. Statistical aspects worth discussing before 1.0

**BernsteinCopula.** I think there are some real issues here. In particular, for empirical Bernstein constructions we currently only issue a warning when `m` does not divide `n`, although this can affect whether the resulting object has exactly uniform margins. Also, the current `logpdf` truncates positive log-densities using `min(log(dens), 0)`, even though a copula density can of course be larger than one.

**EmpiricalCopula.** The finite-sample empirical copula does not have exactly continuous uniform margins, even though `EmpiricalCopula <: Copula`. I do not necessarily think this is wrong as an implementation choice, but I think we should agree on the intended semantics before `Copula` becomes a stable 1.0 abstraction. There also seems to be a small issue in the current atomic `logpdf` implementation.

**Pseudo-observations and ties.** Currently, pseudo-observations resolve ties using ordinal ranks, meaning that tied observations may depend on their order in the input. I think the default tie convention should at least be an intentional statistical decision before 1.0, since this is not a purely computational detail.

**MLE versus pseudo-likelihood.** Maximizing the copula likelihood over the supplied `U` values is fine, but when `U` consists of rank-based pseudo-observations, the interpretation of the covariance matrix and standard errors differs from ordinary maximum likelihood with known margins. Since `CopulaModel` exposes inference functionality, I think we should decide whether we want an explicit distinction between `:mpl` and `:mle`.

**Validity of nested Archimedean copulas.** Currently, the interface allows nested constructions without checking the nesting conditions. This can potentially produce an object typed as `Copula` that is not actually a valid copula. I think we should decide what level of validation belongs in the public constructor.

P.S. I am not very familiar with the current nested implementation, but a few ideas come to mind for a future 1.x release 😄

# 2. API aspects worth discussing before 1.0

I do not think these necessarily need to be implemented now. The important thing is simply to avoid freezing an API that would make them awkward to support later.

**SurvivalCopula vs rotations/reflections.** Currently, `SurvivalCopula(C, flips)` also represents partial coordinate reflections. I wonder whether it would be cleaner to reserve `SurvivalCopula(C)` for the full survival copula and use something like `RotatedCopula` or `ReflectedCopula` for partial reflections.

For example, in the bivariate case I had already considered something like

`RotatedCopula(ClaytonCopula(...), 90)`

while in higher dimensions the more natural terminology would probably be `ReflectedCopula`. I mention this mainly because it seems closer to the standard terminology used in the literature.

**Fixed parameters in fitting.** I do not think this needs to be implemented before 1.0, but I would like the fitting API to remain compatible with something like

`fit(...; fixed=(ν=4.0,))`

later on. In particular, the distinction between free and fixed parameters would affect `coef`, `vcov`, `dof`, AIC/BIC, etc.

**Meaning of fitting methods.** Related to the previous MLE/MPL point, I think we should clearly specify what `:mle` means as part of the stable public API, even if the optimizer itself does not change.

**Empirical copula abstraction.** I think there may also be an API question here. We currently use `EmpiricalCopula` for the Deheuvels-type empirical copula function, although this finite-sample object is not itself a genuine copula on the full unit cube.

Perhaps `EmpiricalCopula` could eventually act as a higher-level interface for empirical copula estimators, while the concrete Deheuvels construction has a more explicit name.

Separately, for the Bernstein issue, it may be worth looking at the empirical checkerboard Bernstein copula (ECBC), which provides a genuine copula for arbitrary polynomial degrees, and also at the broader class of smooth/data-adaptive empirical copula estimators studied by Kojadinovic and Yi (2024).

I would not necessarily implement any of this before 1.0, but I think the current naming and abstraction are worth discussing now.

# 3. Important additions that I do not think should delay 1.0

These all look additive to me and should be naturally compatible with later 1.x releases:

* multivariate `ArchimaxCopula`;
* generic / Khoudraji-type asymmetrization;
* `MixtureCopula` — currently the `Distributions.jl` mixture API does not allow us to fully exploit the copula-specific operations;
* weak tail-dependence coefficients such as `χ̄ₗ` / `χ̄ᵤ`;
* public tools for the Kendall distribution;
* fitting with fixed parameters;
* full likelihood fitting for `SklarDist`;
* generator transformations through outer powers;
* tests for serial independence;
* specialized sampling for nested Archimedean copulas;
* possibly empirical resampling such as Schaake-shuffle-style methods.

So my criterion would essentially be:

**fix statistical errors now, discuss decisions that would be difficult to change after 1.0, and leave additive features for 1.x.**

One small additional point: from the perspective of `VineCopulas.jl`, some of these decisions have direct consequences there, especially rotations/reflections, pair-copula fitting semantics, tie handling, and the fitting extension interface.

Currently, `VineCopulas.jl` also extends a few internal fitting helpers from `Copulas.jl`. Therefore, once we agree on the `Copulas.jl` 1.0 public API, I think it would be useful to review that interaction with downstream packages separately.

lrnv:
Thanks @Santymax98 for the detailed statistical and API review. I split parts 1 and 2 into focused issues so each decision can be discussed, implemented, and closed independently.

## 1. Statistical aspects worth discussing before 1.0

- #465 
- #466 
- #467 
- #468 
- #469

## 2. API aspects worth discussing before 1.0

- #470
- #471
- #472
- #473
- #474 
- #481 

## 3. Other found bugs that i'd like to fix before 1.0

- #477 
- #480

lrnv:
@Santymax98 Thanks again for the detailed pre-1.0 review. The release branch has now been updated from the current `main`, and the draft release notes have been revised to reflect the final API and statistical decisions.

All focused issues extracted from your review have now been addressed.

The resulting fitting boundary is intentionally small: downstream code should use `fit` and `fitting_methods`, while underscored fitting machinery remains internal. Fitted-model uncertainty now belongs to a separate `CopulaInference`, model-selection history to `CopulaSelection`, and ordinary `CopulaModel` state has been minimized.

For VineCopulas.jl, the intended migration path is through behavioral APIs such as `condition` with `cdf`/`quantile`, `params`, and the reflection accessors `basecopula`, `flipmask`, and `flips`. Performance specialization of the public bivariate conditional path is tracked separately in #487 and does not need to delay 1.0.

We will keep this PR in draft until you have had a chance to look at the updated state. Do you see any remaining statistical error, API decision, or other point that should still be addressed before Copulas.jl 1.0?

Santymax98:
Hi @lrnv , thanks a lot for the comments and congratulations on the great work. I am currently migrating the package to become the first downstream package for Copulas.jl, and I will work on releasing it here as soon as possible. Also, if I come across any minor statistical bugs, I will certainly let you know; I am still testing the new version.

lrnv:
A few new issues to deal with before going to 1.0: 

- #490 
- #491
- #492
- #493
- #494
- #495
- #496
- #497
- #498
- #499
- #500
- #501
- #502
- #503
- #505

dcelisgarza:
Hi Oskar,

As you know i've had my eye on Copulas and VineCopulas for a while, and we've spoken over at the defunct discussions but more recently in the [discourse](https://discourse.julialang.org/t/ann-vinecopulas-jl-c-vines-d-vines-and-experimental-r-vines-in-julia/137930).

Would it be cool if I open some feature request issues? Maybe even PRs as I don't want to be too much of a burden.

I've spoken to Santy as well, and will open some issues/PRs over at VineCopulas too. But there are some that would be better served here.

lrnv:
> Hi Oskar,
> 
> As you know i've had my eye on Copulas and VineCopulas for a while, and we've spoken over at the defunct discussions but more recently in the [discourse](https://discourse.julialang.org/t/ann-vinecopulas-jl-c-vines-d-vines-and-experimental-r-vines-in-julia/137930).
> 
> Would it be cool if I open some feature request issues? Maybe even PRs as I don't want to be too much of a burden.
> 
> I've spoken to Santy as well, and will open some issues/PRs over at VineCopulas too. But there are some that would be better served here.

Hey! Of course be my guest ! There's a list right here of tickets, you can edit the list and add you own tickets if you want

Issue #495, opened 2026-09-15
Inference silently regularizes covariance matrices while the documentation promises explicit failure


lrnv:
`Inference.jl:51-70` progressively adds `λI` to the observed information until Cholesky succeeds.

Then around `138-151`, covariance eigenvalues are forced through:

```julia
λ_reg = map(x -> max(x, 1e-12), λ)
```

So a singular or indefinite covariance can silently become a finite covariance estimate.

But `docs/src/manual/fitting_interface.md:247-252` says:

> “There is no generic silent fallback […] if […] fails numerically, `infer` throws.”

Both cannot be the 1.0 contract.

For a statistical API, I would prefer failure over silently manufactured standard errors. Alternatively, `CopulaInference` should expose an explicit regularization diagnostic.

PR #506, opened 2026-09-15
Fix Godambe inference for multivariate rank fits

### LINKED ISSUE #492: `infer` / Godambe is not consistent with multivariate rank-based estimators
In `src/Inference.jl:154-168`, the default for fits using `:itau`, `:irho`, `:ibeta`, or `:iupper` becomes:

```julia
return :godambe
```

However, `_vcov_godambe(... pairwise=false)` builds only **one scalar moment**:

```julia
q = 1
Dα = ForwardDiff.jacobian(
    αv -> [φ(_vcov_copula(...))],
    α,
)
```

For a 3D `GaussianCopula`, for example, the rank-based fit estimates three correlations from **pairwise** dependence measures, whereas `:godambe` uses a single multivariate scalar coefficient.

That produces a sensitivity matrix with rank at most one for multiple parameters.

The code then does:

```julia
DtD = Dα' * Dα
stabilized_inv = inv(DtD + 1e-10I)
```

which turns an underidentified covariance problem into an artificially invertible matrix.

I would not freeze that as the default “analytic” inference path in 1.0.

There is also a reproducibility issue: the Godambe meat is estimated using bootstrap resampling with:

```julia
B = clamp(Int(floor(sqrt(n))), 10, 200)
rng = Random.default_rng()
```

with no way for the user to pass an RNG.

So:

```julia
infer(M)
```

can produce different covariance matrices between runs without a reproducibility hook.

### Recommendation

For 1.0:

* only select Godambe automatically when its estimating equations actually correspond to the estimator used by the fit;
* for unsupported multivariate rank estimators, require an explicit `:bootstrap` or `:jackknife`;
* expose `rng` and resampling count if Godambe retains an internal bootstrap component;
* do not silently solve rank deficiency through `+ 1e-10I`.

Santymax98:
## Summary

Fix the inconsistency between multivariate rank-based estimators and their Godambe covariance calculation.

Closes #492.

## Changes

- use scalar `:godambe` only for supported bivariate rank-matching fits;
- use `:godambe_pairwise` by default for supported multivariate :itau`, `:irho`, and `:ibeta` fits, matching the pairwise moments used by the estimator;
- reject scalar multivariate Godambe inference explicitly;
- check identification through the rank of the sensitivity matrix instead of regularizing an underidentified `D'D`;
- add `rng` and `nresamples` controls to Godambe inference;
- add reproducibility and multivariate regression tests;
- update the inference documentation.

This PR intentionally does not address the more general covariance/Hessian regularization behavior tracked in #495.


Santymax98:
@lrnv need re run please. Thanks

Issue #502, opened 2026-09-15
`TCopula`: add `:itau` (`ρ = sinpi(τ̂/2)`, `ν` by profile likelihood at fixed `ρ`) beside `:itau_irho`, and extend the correlation half to `d > 2`


dcelisgarza:
Written against `main` at 67dea70 (the 1.0 branch).

Companion to my Gaussian rank-inversion issue #501.

## Today

`TCopula` advertises `(:mle, :itau_irho)` in `d = 2` and `(:mle,)` above (`src/EllipticalCopulas/TCopula.jl:373`). `:itau_irho` (`:341`) takes the closed form `ρ = sinpi(τ̂/2)`, then identifies `ν` from Spearman's ρ by bisection on `|ρ_S(ν; ρ)| = |ρ̂_S|`. Two gaps for a downstream vine fitter:

1. **No `d > 2`.** The correlation half, `R = sinpi.(τ̂ ./ 2)` pairwise plus a correlation repair, is exactly as closed form in `d > 2` as in `d = 2`; only the `ν` half needs a one-dimensional search, which the `:mle` profile (`:305`) already performs.
2. **The estimator that vine libraries call `itau`** for the Student-t is `ρ` from τ and **`ν` from the likelihood with `ρ` fixed**, a one-dimensional profile. Your `:mle` is that profile over `λ = 1/ν` with `ρ` re-estimated at every `ν` through `_fit_t_corr_given_nu` (`:264`); `:itau` is the same loop with the correlation held at the τ inversion, so the inner optimisation disappears. A downstream selector that asks Copulas.jl for `:itau` per family such as VineCopulas.jl's `pair_method = :itau`, currently gets an `ArgumentError` for the Student-t and silently drops it from the candidate set.

## What I measured with some clanker help on Copulas 0.1.43, `d = 2`, 50 seeds each

I expected `:itau_irho` to identify `ν` poorly from the τ/ρ_S gap, so I had a look. It turns out the two rank estimates are strongly correlated on the same sample, so their gap is much less noisy than either method's estimate. Quantiles (5 %, 50 %, 95 %) of `ν̂`, and mean time per fit:

| truth | `n` | profile `:itau` | `:itau_irho` |
|---|---|---|---|
| `ν = 4, ρ = 0.6` | 1000 | 3.08 / 3.83 / 4.92, 0.08 s | 2.89 / 3.97 / 4.88, 0.39 s |
[...]

PR #507, opened 2026-09-15
Closed-form rank inversions for the elliptical copulas

### LINKED ISSUE #501: `fit(GaussianCopula, U; method = :itau)` runs a Nelder–Mead search for a quantity that is closed form — use `sinpi(τ̂/2)` (and `2 sinpi(ρ̂_S/6)` for `:irho`), pairwise with a correlation repair for `d > 2`
Hi Oskar,

I'm building on Copulas.jl (through VineCopulas.jl) in PortfolioOptimisers.jl and I'm writing this against `main` at 67dea70 (the 1.0 branch), since the fitting surface is being settled now, I thought it best to do this now.

## Today

`GaussianCopula` advertises `(:mle, :itau, :irho, :ibeta)` (`src/EllipticalCopulas/GaussianCopula.jl:223`) but only implements `:mle` (`:165`). `:itau` and `:irho` therefore fall to the generic rank routine at `src/Fitting.jl:237–250`: build the example copula, unbound its parameters, and run `Optim.NelderMead` on `sum(abs2, τ̂ .- τ(C))` through the partial-correlation reparameterisation.

For `d = 2` the answer is closed form: `τ = (2/π) asin ρ` (your `τ(C::GaussianCopula{2})` at `:107`), so `ρ = sinpi(τ̂/2)`; likewise `ρ_S = (6/π) asin(ρ/2)`, so `ρ = 2 sinpi(ρ̂_S/6)`. The Nelder–Mead route returns a value about **1e-4 off** the closed form (measured 8.4e-5 on a 2000-observation sample from `TCopula(4, 0.6)` on Copulas 0.1.43; the generic routine is unchanged on `main`) at the cost of an optimiser run, and the same inaccuracy propagates into every downstream τ-inversion (VineCopulas.jl's `pair_method = :itau` goes through this method for the Gaussian family).

```julia
using Copulas, StatsBase
U = rand(GaussianCopula(2, 0.6), 2000)
τ̂ = corkendall(U')[1, 2]
Distributions.params(fit(GaussianCopula, U; method = :itau)).Σ[1, 2] - sinpi(τ̂ / 2)   # ≈ -8e-5, not 0
```

For `d > 2` the same holds pairwise: `R̂_ij = sinpi(τ̂_ij/2)` is the standard estimator, and it needs only a projection to the nearest correlation matrix when the pairwise matrix is not positive definite, which `_score_corr_start` (`src/EllipticalCopula.jl:169`) already does by shrinkage toward the identity.

## Proposal

Add the two methods beside the Gaussian `:mle`, keeping `:ibeta` on the generic route:

```julia
function _fit(::Type{<:GaussianCopula}, U, ::Val{:itau})
    T = float(eltype(U))
    τ̂ = StatsBase.corkendall(U')
    R = sinpi.(τ̂ ./ 2)
    return GaussianCopula(_nearest_correlation(R))
end
function _fit(::Type{<:GaussianCopula}, U, ::Val{:irho})
    ρ̂ = StatsBase.corspearman(U')
    R = 2 .* sinpi.(ρ̂ ./ 6)
    return GaussianCopula(_nearest_correlation(R))
end
```

where `_nearest_correlation(R)` symmetrises, sets the diagonal to one, returns `R` when it is positive definite, and otherwise applies the same shrinkage loop as `_score_corr_start` (or Higham's alternating projections if you would rather have the true nearest matrix; the shrinkage is what the package already trusts). For `d = 2` no projection is ever needed and the method is exact.

The Student-t copula shares `τ = (2/π) asin ρ`, so the same `R` is the `:itau` correlation for `TCopula`; I'm proposing its `ν` half in a separate issue so this one stays a one-family change.

## Tests

- `d = 2`: `fit(GaussianCopula, U; method = :itau)` returns `Σ[1,2] == sinpi(corkendall(U')[1,2]/2)` to `1e-15`, and `:irho` the Spearman analogue.
- `d = 4` on a sample from a known correlation matrix, `n = 5000`: the `:itau` estimate is within `0.03` of the truth entrywise and is positive definite.
- A pairwise matrix that is not positive definite (three variables with `τ̂ = (0.9, 0.9, −0.9)`) still returns a valid `GaussianCopula`.
- Round trip: `τ(fit(GaussianCopula, U; method = :itau)) ≈ corkendall(U')[1,2]` at `d = 2`.
- The generic rank routine is still reached for a family without a specialised method (a regression guard).

I'm happy to send this as a PR; it is ~30 lines including the projection helper.

dcelisgarza:
# Closed-form rank inversions for `GaussianCopula`, and a `:itau` for `TCopula` in every dimension

Written against `main` (67dea70). Fixes #501 and #502 .

Small, self-contained, no new dependency. I used opus 5 to check the code, tests, and ensure it fits in with the style of Copulas.jl (unicode names are the bane of my existence lol).

## What was wrong

`GaussianCopula` advertises `(:mle, :itau, :irho, :ibeta)` but only implements `:mle`, so `:itau` and `:irho` fall to the generic rank routine in `src/Fitting.jl`: build the example copula, unbound its parameters, and run `Optim.NelderMead` on `sum(abs2, τ̂ .- τ(C))` through the partial-correlation reparameterisation. For a Gaussian pair `τ = (2/π) asin ρ` and `ρ_S = (6/π) asin(ρ/2)`, so the answer is `ρ = sinpi(τ̂/2)` and `ρ = 2 sinpi(ρ̂_S/6)`; the search lands about 1e-4 away from it (8.4e-5 on a 2000-observation bivariate sample) at the cost of an optimiser run, and the inaccuracy propagates into every downstream τ-inversion built on this method.

```julia
using Copulas, StatsBase
U = rand(GaussianCopula(2, 0.6), 2000)
τ̂ = corkendall(U')[1, 2]
fit(GaussianCopula, U; method=:itau).Σ[1, 2] - sinpi(τ̂ / 2)   # ≈ -8e-5 on main, 0.0 here
```

`TCopula` advertised `(:mle, :itau_irho)` in dimension 2 and `(:mle,)` above, so a caller asking for `:itau` per family, a vine fitter selecting among families with one rank estimator, for instance, got `ArgumentError: Method 'itau' not available for TCopula` and had to drop the Student copula from its candidate set.

## What this does

**`GaussianCopula`**: `:itau` and `:irho` become the closed forms above, applied entrywise to `StatsBase.corkendall(U')` / `corspearman(U')`, in every dimension. `:ibeta` stays on the generic route. In dimension 2 the result is exact to the last bit (`Copulas.τ(fit(GaussianCopula, U; method=:itau)) == corkendall(U')[1, 2]` up to rounding of `asin ∘ sin`).

**`TCopula`**: a new `:itau`, advertised in every dimension. Kendall's τ of an elliptical copula depends on the correlation alone, so the correlation matrix is the same pairwise closed form as for the Gaussian copula; the degrees of freedom are then the maximiser of the likelihood with that correlation held fixed. That is the profile over `λ = 1/ν` that `:mle` already runs (`_t_profile_upper` bracket, Brent, the Gaussian endpoint `λ = 0` with the same tolerance rule), minus the inner correlation optimisation at every `ν`, so it is the conventional "itau" for the Student copula and it shares every helper with `:mle`. `:itau_irho` is unchanged and stays bivariate.

On a 5000-observation bivariate sample from `TCopula(4, [1 0.6; 0.6 1])`: `:itau` gives `ν̂ = 3.83`, `:mle` gives `3.83`, `:itau_irho` gives `3.60`; the `:itau` log-likelihood is 2.8e-4 below the MLE's. In dimension 3 with `n = 2000` the fit takes ~0.4 s.

**`_nearest_correlation`**: a pairwise matrix of inverted coefficients need not be positive definite above dimension 2 (e.g. pairwise τ̂ of `(0.9, 0.9, −0.9)`). The helper symmetrises, sets the diagonal to one, and when the smallest eigenvalue is below `sqrt(eps)` shrinks toward the identity by the single `λ = (δ − λ_min)/(1 − λ_min)` that puts it at `δ`, the eigenvalues of `(1 − λ)R + λI` are `(1 − λ)λ_i + λ`. `_score_corr_start` now delegates to it. Its previous loop stepped `λ` through `sqrt(eps), 10 sqrt(eps), …` and returned the **identity** when `λ = 0.15` was not enough, which for the triple above threw the whole matrix away; the new form returns `[1 0.5 0.5; 0.5 1 −0.5; 0.5 −0.5 1]`. A sample correlation of normal scores is always positive semidefinite, so the likelihood starting point only differs on the singular `n ≤ d` case, where the first step of the old loop and the closed form coincide (`λ = sqrt(eps)`). Happy to keep the old loop in `_score_corr_start` untouched if you would rather not touch the MLE start at all, it is a one-line revert.

I did not touch the `ν = Inf` endpoint rule. Both `:mle` and the new `:itau` return a very large finite `ν` (≈ 7e6) rather than `Inf` on a Gaussian sample, because `quantile(TDist(ν), u)` is noisy at that scale and the finite profile value ends up a few 1e-5 above the exact Gaussian log-likelihood, outside the `100 eps` tolerance. That is pre-existing and separate from this change; I can look at it in a follow-up if you want.

## Tests

In `test/operations/fitting.jl`:

- *Gaussian rank inversions are closed form*: `d = 2` exactness to the bit for `:itau` and `:irho`, round trip through `Copulas.τ`/`Copulas.ρ` to 1e-14; `d = 4` on a known matrix, both methods within 0.03 entrywise and positive definite, `:itau` equal to `sinpi.(corkendall(U') ./ 2)` exactly; the `(0.9, 0.9, −0.9)` triple is repaired to a positive-definite matrix with unit diagonal that keeps the sign pattern and does not collapse to the identity; an already valid matrix passes through unchanged.
- *Student Kendall inversion profiles degrees of freedom*: `d = 2`, `n = 5000` from `TCopula(4, 0.6)`: the correlation is exactly the τ inversion, `ν̂ ∈ [3, 6]`, the log-likelihood is bounded above by the MLE's and below by its own Gaussian endpoint, `ν̂` within 10 % of the MLE's; `d = 3` through `fit(CopulaModel, …)`, method reported as `:itau`, pairwise-exact correlation, positive definite, `ν̂ ∈ [3, 6]`; `:itau` advertised in `d = 3`, `:itau_irho` not.
- The existing *one execution per fitting engine* sweep now executes both new routes, since they are new `_fit` methods.

Locally: `test/operations/fitting.jl` (1386 passed), `test/correctness/elliptical.jl`, `test/operations/fitting_selection.jl`, `test/operations/inference.jl`, `test/Aqua.jl` all pass on Julia 1.13.

## Docs and benchmarks

The estimating-principle list in `docs/src/manual/fitting_interface.md` says what `:itau`/`:irho` do for the elliptical families; the `GaussianCopula` and `TCopula` sections of `docs/src/bestiary/elliptical.md` each gain a paragraph and a runnable `@example`. `benchmark/benchmarks.jl` gains `gaussian_itau` and `student_itau` rows next to the existing `gaussian_mle` and `student_rank_matching`.


dcelisgarza:
Looks like we found a nice little security bypass in github's PR API. Glad CI now works.

PR #508, opened 2026-09-16
Condition on intervals, and treat a discrete observation as one

### LINKED ISSUE #491: `SklarDist` currently claims support for discrete and mixed margins, but the generic implementation is mathematically wrong
This is the second major architectural issue.

`src/SklarDist.jl:67` defines:

```julia
struct SklarDist{CT,TplMargins} <: Distributions.ContinuousMultivariateDistribution
```

while the constructor accepts arbitrary `UnivariateDistribution`s.

More importantly, `_logpdf`, around lines `141-155`, always applies the continuous Sklar density formula:

```julia
sum(logpdf(margin_i, x_i)) + logpdf(C, F.(x))
```

That factorization is not valid for discrete margins.

The documentation already acknowledges this in `SklarDist.jl:29-34`: with discrete or mixed margins, multiplying marginal masses/densities by the copula density generally does not produce the correct joint mass.

But the generic code still performs exactly that computation.

A very simple counterexample is:

```julia
S = SklarDist(
    FGMCopula{2}(1.0),
    (Bernoulli(0.5), Bernoulli(0.5)),
)
```

The correct joint mass at `(0,0)` is

$$
C(0.5,0.5)=0.3125.
$$

The current generic `_logpdf` instead gives

$$
0.5 \times 0.5 \times c(0.5,0.5)=0.25.
$$

So this is not merely a documentation mismatch: `pdf` returns the wrong probability.

There is a specialization for `EmpiricalCopula` in `EmpiricalCopula.jl:68-104`, which explains why some current Bernoulli tests can pass, but the general problem remains for ordinary copulas.

The same conceptual issue appears in conditioning. `Conditioning.jl:389-399` transforms an observed discrete margin using only:

```julia
u = cdf(margin, x)
```

whereas the documentation correctly explains that a discrete observation corresponds to a **latent interval**, not simply to the upper endpoint `F(x)`.

The same applies to Rosenblatt transforms, which currently map margins through `cdf` and then apply the copula transform.

### Recommendation for 1.0

Rather than reducing the API to continuous margins, I would now recommend **finishing the discrete and mixed-margin implementation before 1.0**.

That means implementing the actual Sklar probabilities for atomic coordinates.

For fully discrete margins, the joint mass should be obtained through finite differences of the copula CDF over the corresponding probability rectangle:

$$
P(X=x)
=
\Delta_1 \cdots \Delta_d
C(F_1(x_1),\ldots,F_d(x_d)),
$$

with lower bounds

$$
F_i(x_i^-).
$$

For mixed continuous/discrete margins, the implementation should combine:

* derivatives with respect to continuous coordinates;
* finite differences with respect to discrete coordinates.

In other words, for continuous coordinates \(C\) and discrete coordinates \(D\), the likelihood contribution should be the appropriate mixed partial / rectangle difference of the copula, multiplied only by the continuous marginal densities.

Conditioning must follow the same semantics: observing a discrete margin means conditioning on

$$
F(x^-) < U \le F(x),
$$

not on a single latent value.

Before calling this 1.0-ready, I would therefore want:

* correct `pdf/logpdf` for fully discrete margins;
* correct mixed discrete/continuous likelihoods;
* correct conditioning semantics for discrete observations;
* explicit behavior for Rosenblatt transforms when atoms are present;
* tests against hand-computable 2D examples such as Bernoulli × Bernoulli under FGM;
* consistency tests showing that the continuous implementation is recovered when all margins are continuous.

This is more work than restricting the API, but if discrete/mixed `SklarDist` is meant to be a real 1.0 feature, this is the right time to implement it properly rather than freezing an incorrect approximation.

The current supertype `ContinuousMultivariateDistribution` should also be reconsidered if discrete and mixed distributions are genuinely supported.

### LINKED ISSUE #505: `condition` on an interval: `U_J ∈ ∏[a_j, b_j]` as a ratio of box measures, mixed with point conditioning, and the discrete-margin case it makes correct
Hi Oskar,

This is me being *very* cheeky, especially after the our very illustrative exchange on discourse over the weekend. But in my defense, i think this would be a great addition. I also don't know if you can deal with types other than floats, so I'm proposing using `eltype` instead of `float` even though there's a type promotion that bounds to at least Float64 (maybe worth removing it in the actual implementation).

---

Written against `main` at 67dea70 (the 1.0 branch). This is a conditioning ask, separate from my two fitting ones; it is "small" because the primitives are already in the package.

## What I need

I'm building a stress-scenario generator on your conditioning framework. A stress is rarely a point: it is "asset 3 in its bottom decile", `U_3 ∈ [0, 0.1]`, or a joint box on two or three coordinates, and the question is the law of the other coordinates given that event. Today `condition(C, js, u_js)` (`src/Conditioning.jl:382`) computes the regular conditional at a point through mixed partials, and the manual is explicit that this is the only version it computes: `docs/src/manual/conditioning_and_subsetting.md:51-53` says a discrete margin corresponds to *an interval of latent uniforms* and "the formula above does not establish correct conditioning for that case". The interval version is the same object your `measure` (`src/Copula.jl:503`) already builds, so I'd like to propose it as a second form of `condition`, and I can send it as a PR.

Substituting the interval's midpoint into the point conditional is not a substitute: on `ClaytonCopula(3, 2.0)` with `U_3 ∈ [0, 0.1]`, the CDF of `U_1 | U_3 = 0.05` and the CDF of `U_1 | U_3 ∈ [0, 0.1]` differ by up to 0.093 (sup over `u ∈ 0.05:0.05:0.95`).

## The law, in the package's own primitives

Partition the conditioned set into point coordinates `P` (values `u_P`) and interval coordinates `B` (box `∏_{j ∈ B} [a_j, b_j]`), with `I` free. The conditional CDF of `U_I` is the inclusion–exclusion sum over the `2^|B|` corners of the box of the point primitive you already have, with the interval coordinates placed among the *free* ones of `_partial_cdf` (`src/Conditioning.jl:52`):

```math
H_{I \mid P, B}(u_I) = \frac{\sum_{c \in \text{corners}(B)} \textrm{sgn}(c)\, \partial_{u_P} C(u_I, u_P, c)}{\sum_{c \in \text{corners}(B)} \textrm{sgn}(c)\, \partial_{u_P} C(\mathbf 1_I, u_P, c)},
```

where

```math
\textrm{sgn}(c) = (−1)^{\{j : c_j = a_j\}} \,.
```

With `P = ∅` the derivatives disappear and both sides are C-volumes: numerator `measure(C, (0_I, a_B), (u_I, b_B))`, denominator the box's measure under the `B`-margin, which is `b − a` when `|B| = 1` (a copula margin is uniform; `subsetdims(C, (j,))` returns `Uniform()`, so the sketch handles that case without calling `measure` on it). With `B = ∅` it is your current formula. The denominator is a probability, not a density, so the interval form has no zero-density failure mode: it is well defined wherever the box has positive measure, and needs no "choice of version".

The conditional density on `u_I` is the `|I|`-th mixed partial of the same sum, i.e. `_partial_cdf` differentiated with respect to `(P..., I...)` at the corners, so `logpdf`, `quantile` and sequential sampling all reduce to the one primitive.

**Measured on `main` at 67dea70**, against rejection samples (`rand(C, 2_000_000)`, keep the rows in the box):

| case | check | max abs error |
|---|---|---|
| `ClaytonCopula(3, 2.0)`, `U_3 ∈ [0, 0.1]` | joint CDF of `(U_1, U_2)` at 25 grid points, 200 116 rows kept | 0.0013 |
| same | marginal distortion of `U_1`, 19 points | 0.0019 |
| same, mixed: `U_2 = 0.7` and `U_3 ∈ [0, 0.1]` | CDF of `U_1` against `rand(condition(C, (2,), (0.7,)), 10^6)` rejected on `U_3`, 2 857 rows kept | 0.0052 |
| `GaussianCopula([1 .6 .3; .6 1 .5; .3 .5 1])`, `U_2 ≤ 0.2, U_3 ≤ 0.2` | acceptance 0.08691 vs `measure` 0.08718; CDF of `U_1`, 19 points, 173 814 rows kept | 0.0025 |

All at Monte Carlo resolution. Per evaluation on the Clayton: 0.8 µs for the interval CDF (two `measure` calls on a closed-form CDF), 2.1 µs for the mixed case, against 157 µs for constructing and evaluating the existing point conditional through the AD path.

## Proposed API

One additional method per entry point, no new types in the public surface:

```julia
# U_I | U_js ∈ ∏ [lo_k, hi_k].  A coordinate with lo_k == hi_k is conditioned on that point.
condition(C::Copula, js, lo_js, hi_js)
condition(X::SklarDist, js, xlo_js, xhi_js)     # intervals on the original scale, mapped through the margins
```

Return shape follows `condition(C, js, u_js)` exactly: a `Distortion` when one coordinate is free, otherwise a `SklarDist` of the conditional copula and the distortions. The "`lo == hi` is a point" convention makes the four-argument form a strict generalisation, `condition(C, js, u, u)` is `condition(C, js, u)`, and lets one call mix a fixed value with a box. If you'd rather keep one signature, the same thing reads as `condition(C, js, u_js)` with an entry of `u_js` allowed to be a 2-tuple `(a, b)`; `_process_tuples` (`src/Conditioning.jl:54`) is where either form would be normalised. Validation: `0 ≤ lo ≤ hi ≤ 1`, and an `ArgumentError` when a box has zero measure under the copula.

**What it also buys you:** the discrete-margin caveat in the manual goes away. For a discrete `X.m[j]`, observing `X_j = x_j` is `U_j ∈ [F_j(x_j) − p_j(x_j), F_j(x_j)]`, so `condition(X::SklarDist, js, x_js)` can dispatch a `DiscreteUnivariateDistribution` margin to the interval form and be correct rather than documented as not. I'd put that in the same PR if you want it, or leave it as a follow-up.

## Implementation sketch

One internal primitive, generalising `_partial_cdf` to a box:

```julia
# Inclusion–exclusion over the 2^q corners of the box on coordinates `bs`, of the point
# primitive with those coordinates placed among the free ones. q == 0 is `_partial_cdf`.
function _box_partial_cdf(C, is, ps, bs::NTuple{q,Int}, uᵢₛ, uₚₛ, lo, hi) where {q}
    q == 0 && return _partial_cdf(C, is, ps, uᵢₛ, uₚₛ)
    T = promote_type(eltype(uᵢₛ), eltype(uₚₛ), eltype(lo), eltype(hi), Float64)
    r = zero(T)
    for mask in 0:(1 << q) - 1
        corner = ntuple(k -> (mask >> (k - 1)) & 1 == 1 ? hi[k] : lo[k], q)
        sgn = isodd(q - count_ones(mask)) ? -one(T) : one(T)
        r += sgn * _partial_cdf(C, (is..., bs...), ps, (uᵢₛ..., corner...), uₚₛ)
    end
    return r
end
```

(`measure` at `Copula.jl:503` is this with `ps = ()` and a Gray code; I kept the mask loop so the sign is explicit, and the two agree on every corner set. If you prefer, `measure` can be rewritten on top of it.)

Then `DistortionFromCop` (`:128`) and `ConditionalCopula` (`:210`) each gain three fields: `bs::NTuple{q,Int}`, `lo::NTuple{q,T}`, `hi::NTuple{q,T}`, defaulting to empty tuples, so every existing code path is the `q = 0` case with an unchanged result:

```julia
# DistortionFromCop: numerator and denominator become box sums
Distributions.cdf(d, u)   → _box_partial_cdf(d.C, (d.i,), d.ps, d.bs, (T(u),), d.uₚₛ, d.lo, d.hi) / d.den
d.den                     = _box_partial_cdf(d.C, (), d.ps, d.bs, (), d.uₚₛ, d.lo, d.hi)   # u_I = 1
Distributions.logpdf(d,u) → log(_box_partial_cdf(d.C, (), (d.ps..., d.i), d.bs, (), (d.uₚₛ..., u), d.lo, d.hi)) - log(d.den)

# ConditionalCopula
_cdf(CC, v)               → _box_partial_cdf(CC.C, CC.is, CC.ps, CC.bs, uI, CC.uₚₛ, CC.lo, CC.hi) / CC.den
_logpdf(CC, v)            → the q == 0 branch keeps `logpdf(CC.C, u) - CC.logden` (:273); q > 0 uses
                            log(_box_partial_cdf(CC.C, (), (CC.ps..., CC.is...), CC.bs, (), (CC.uₚₛ..., uI...), CC.lo, CC.hi)) - CC.logden
_rand!(rng, CC, A)        → the sequential loop at :282 unchanged, except each step's distortion carries
                            (CC.bs, CC.lo, CC.hi) alongside the growing point set
```

And the entry point:

```julia
function condition(C::Copula{D}, js::NTuple{p,Int}, lo::NTuple{p,<:Real}, hi::NTuple{p,<:Real}) where {D,p}
    all(0 .<= lo .<= hi .<= 1) || throw(ArgumentError("interval bounds must satisfy 0 ≤ lo ≤ hi ≤ 1"))
    pt = ntuple(k -> lo[k] == hi[k], p)
    ps = Tuple(js[k] for k in 1:p if pt[k]);   uₚₛ = Tuple(eltype(lo[k]) for k in 1:p if pt[k])
    bs = Tuple(js[k] for k in 1:p if !pt[k]);  lob = Tuple(eltype(lo[k]) for k in 1:p if !pt[k])
                                               hib = Tuple(eltype(hi[k]) for k in 1:p if !pt[k])
    isempty(bs) && return condition(C, ps, uₚₛ)
    is = Tuple(setdiff(1:D, js))
    p == D - 1 && return DistortionFromCop(C, ps, uₚₛ, is[1]; bs, lo = lob, hi = hib)
    CC = ConditionalCopula(C, ps, uₚₛ; bs, lo = lob, hi = hib)
    return SklarDist(CC, CC.distortions)
end

function condition(X::SklarDist{<:Copula{D}}, js::NTuple{p,Int}, xlo::NTuple{p,<:Real}, xhi::NTuple{p,<:Real}) where {D,p}
    lo = ntuple(k -> Distributions.cdf(X.m[js[k]], xlo[k]), p)
    hi = ntuple(k -> Distributions.cdf(X.m[js[k]], xhi[k]), p)
    ... # then as the point SklarDist method at :390, pushing the distortions through the free margins
end
```

Family fast paths are not needed for correctness: the Gaussian point conditional is Gaussian (`GaussianCopula.jl:111`), but the Gaussian *interval* conditional is a truncated-normal mixture with no closed form, so the generic route is the right default there too, and its `cdf` already goes through `MvNormalCDF`. Families with a closed-form CDF (Archimedean, extreme-value) get an interval conditional that is cheaper than their point conditional, since no derivative is taken.

## Tests

- **Reduction**: `condition(C, js, u, u)` has the same `cdf`/`logpdf`/`quantile` as `condition(C, js, u)` to `1e-12`, on Gaussian, Clayton and Gumbel at `D = 3`.
- **Rejection parity**: the four rows of the table above as assertions at tolerance `5e-3` with a seeded `2_000_000`-row sample (the mixed row at `2e-2`, its acceptance is 0.3 %).
- **Full box is the margin**: `condition(C, js, zeros(p), ones(p))` has the law of `subsetdims(C, is)`, `cdf` at 20 random points to `1e-12`.
- **Density integrates**: `pdf` of the `D − 1 = 1` distortion integrates to 1 over `[0, 1]` (`QuadGK`), and `quantile ∘ cdf` round-trips.
- **Sampling**: `rand(condition(C, js, lo, hi), 10^5)` on `D = 4`, `|B| = 2`, has the same pairwise Kendall τ as the rejection sample, tolerance `1e-2`.
- **Discrete margin** (if the follow-up is taken): `X = SklarDist(ClaytonCopula(2, 2.0), (Normal(), Poisson(3.0)))`; `condition(X, 2, 2)` against the ECDF of `X_1` over rows of `rand(X, 10^6)` with `X_2 == 2`, tolerance `5e-3`.
- **Errors**: `lo > hi`, bounds outside `[0, 1]`, and a zero-measure box each throw an `ArgumentError`.

Happy to send this as a PR, the primitive is 15 lines, the two structs take three fields each, and I have the fixtures and the rejection oracles from the measurements above.

dcelisgarza:
Fixes #505, fixes #491. Written against `main` at faa02393.

## What this does

`condition` gains a box form, and a discrete observation of a `SklarDist` becomes the latent interval it is. Both sit on one internal primitive, `_box_partial_cdf(C, is, ps, bs, uis, ups, lo, hi)`: the inclusion–exclusion sum over the `2^q` corners of the box on `bs` of the existing `_partial_cdf`, with the box coordinates placed among the free ones. `q == 0` is `_partial_cdf`, `ps == ()` is the C-volume that `measure` computes, and a family that specialises `_partial_cdf` (extreme-value, nested Archimedean) speeds up everything below without further work.

### 1. Interval conditioning (#505)

```julia
condition(C::Copula, js, lo_js, hi_js)          # U_I | U_js ∈ ∏ [lo_k, hi_k]
condition(X::SklarDist, js, xlo_js, xhi_js)     # intervals on the original scale, mapped through the margins
```

A coordinate with `lo == hi` is conditioned on that point, so `condition(C, js, u, u)` is `condition(C, js, u)` and one call mixes fixed values with intervals. The return shape follows the point form: a univariate distribution when one coordinate is free, otherwise a `SklarDist` of the conditional copula and the distortions. `DistortionFromCop{TC,p,q,T}` and `ConditionalCopula{d,D,p,q,T,TDs}` carry `bs`, `lo`, `hi`, and `q = 0` is the unchanged point path, so the family `distortion`/`conditional_copula` hooks are used exactly when every conditioned coordinate is a point. The denominator of the box form is a probability, so it is defined wherever the box has positive probability; a zero-probability box throws an `ArgumentError`, as do `lo > hi`, bounds outside `[0, 1]`, and repeated indices.

I didn't like the four-argument signature much either, but I could not find a better one that does not put a tuple-of-tuples into `u_js`; it is a strict generalisation, so nothing existing moves. I also thought about extending the type boundary to vectors and checking their length as that may be less compiler heavy for large numbers of variables. I think the limit was 4 (?), when tuples started allocating/being slower, but I don't know the state of that nowadays.

### 2. Sklar likelihood with atoms (#491)

`SklarDist` `logpdf` splits the margins by `value_support`: the mixed partial in the continuous coordinates, differenced over `(F_j(x_j⁻), F_j(x_j)]` in each discrete one, times the continuous marginal densities only. All-continuous models take the old code path unchanged (byte-identical result, tested). The left limit is `cdf(m, x) - pdf(m, x)`: `cdf(Poisson(3), prevfloat(2.0))` returns `cdf(…, 2)`, not `cdf(…, 1)`, so `prevfloat` is not a left limit for every family (the `EmpiricalCopula` Sklar method uses it; I left that alone since its test fixture is Bernoulli, where it happens to work).

Measured: Bernoulli × Bernoulli under `FGMCopula{2}(1.0)` gives `pdf(S, [0, 0]) = 0.3125 = C(½, ½)` and the four cells sum to one; Normal × Poisson under `ClaytonCopula(2, 2.0)` integrates out to the Poisson pmf to `6e-16` for `k = 0…8`, and `E[X₁ | X₂ = 2]` from the density matches a 10⁶-row sample to `2e-3`.

### 3. Conditioning on a discrete observation

`condition(X, js, x_js)` conditions a discrete margin on its latent interval and a continuous one on its point, in one call; an observation of zero probability throws. `DistortedDist` `logpdf` returns the conditional mass `H(F(t)) − H(F(t⁻))` when the *free* margin is discrete, so `pdf` of `X₁ | X₂ = 0.4` with `X₁ ~ Poisson` sums to one and cumulates to its `cdf`. The `PartitionedDistributions` extension inherits this through `condition(X, js, x_js)` (tested); an interval selector has no counterpart in that package's `conditional(dist, x, keep)`, so I did not add one there.

### 4. Rosenblatt with atoms

Two rules, both measured before I picked them. A discrete predecessor conditions on its interval, and a discrete coordinate itself takes the distributional transform of its conditional law, `S_i = H(F(x⁻)) + V·(H(F(x)) − H(F(x⁻)))`, `V ~ U(0,1)`. The obvious shortcut, randomise the atom and condition the successors point-wise on that latent value, is wrong: the randomised latent is independent of everything, whereas `U_i` on the atom's interval is not. On `SklarDist(ClaytonCopula(2, 2.0), (Poisson(3.0), Normal()))`, the max deviation of the transformed pair's joint CDF from independent uniforms is 0.17 today, 0.024 with the shortcut, 0.0017 with the interval (5·10⁵ rows), and the test pins that ordering. So `rosenblatt(rng, X, x)` is random on atoms (`rosenblatt(X, x)` draws from the default rng, and is unchanged for continuous margins), and `inverse_rosenblatt(X, s)` needs no randomness: step `i` is `Q_i(H⁻¹(s_i))`, the generalised inverse `DistortedDist` already computes. `inverse_rosenblatt(X, rosenblatt(rng, X, x)) == x` on mixed fixtures, and the inverse reproduces `P(X₁ = k, X₂ ≤ t)` to `5e-3`.

## A Gumbel defect this exposed

The generic AD path places every free coordinate at 1, and `log(-log(1))` is a `NaN` dual (`log(Dual(0, 0))` has partial `0/0`), so `Copulas._partial_cdf(GumbelCopula(3, 1.6), (2, 3), (1,), (0.4, 1.0), (0.7,))` was `NaN` on `main`, and with it the generic `DistortionFromCop` of any Gumbel in `D ≥ 3`. The Archimedean family hook hides this on the point path; the box path has no family hook, so a Gumbel with a discrete margin hit it in `rosenblatt`. `ϕ⁻¹` of the Gumbel and BB3 generators is now `(-log t)^θ` (equal to an ulp, and AD-safe at `t = 1`), and the Gumbel `_archimedean_cdf` skips the coordinates at 1, which contribute nothing to its log-space sum. A regression test pins the generic partial against the family path.

## Types

Everything is derived from the arguments with `promote_type` / `promote`; nothing is cast to `Float64`, and the `BigFloat` box path is tested (`den`, bounds, `cdf`, `logpdf` all `BigFloat`). I dropped the `float(...)` in `DistortionFromCop`'s `cdf` for the same reason; the result type is `promote_type(typeof(u), typeof(den))`.

## Tests

- Reduction: `condition(C, js, u, u)` has the same `cdf`/`logpdf`/`quantile` as `condition(C, js, u)` to `1e-12` on Gaussian, Clayton and Gumbel at `D = 3`, and the joint form at `D − p = 2`.
- Rejection parity on seeded 10⁶-row samples: `ClaytonCopula(3, 2.0)` with `U₃ ∈ [0, 0.1]` (joint CDF at 25 points, marginal at 10), the mixed `U₂ = 0.7, U₃ ∈ [0, 0.1]`, the Gaussian `U₂, U₃ ≤ 0.2` (whose `den` equals `measure` to `1e-12`), and a 4-D `|B| = 2` case by Kendall's τ of `rand` against the rejection sample and by `hcubature` of the joint density.
- Full box is the margin; the mixed density integrates to one; `quantile ∘ cdf` round-trips; the midpoint substitute is off by more than 0.05.
- Errors: `lo > hi`, bounds outside `[0, 1]`, repeated indices, length mismatch, a zero-probability box (`MCopula(3)` off the diagonal), `xlo > xhi`, a zero-probability discrete observation.
- Discrete: FGM Bernoulli cells, Normal × Poisson marginalisation, three discrete margins under a Gaussian sum to one, `X₁ | X₂ = 2` against the rejection sample (and the endpoint `F(2)` is off by more than 0.05), an original-scale interval on a discrete margin, a discrete free margin's pmf, a mixed observed set in 3-D.
- Rosenblatt: independent uniforms on the forward transform, the point-wise shortcut fails, the inverse matches the joint law, round trips on two mixed fixtures, and continuous `SklarDist`s keep the deterministic copula route.

## Docs

The manual's "the formula above does not establish correct conditioning for that case" paragraph is replaced by *Conditioning on an interval* and *Discrete margins* sections with the formula, an example next to the midpoint conditional, and the `2^k` cost note; `condition`, `rosenblatt`, `inverse_rosenblatt` and `SklarDist` docstrings, the public API table, and the developer guide's conditioning section say the same.

## Not in this PR

- The `ContinuousMultivariateDistribution` supertype of `SklarDist` (#491's last point). Orthogonal; your call.
- A deterministic mid-distributional-transform variant of `rosenblatt` (`V = ½`, what a checkerboard goodness-of-fit wants). One keyword if you want it; I defaulted to the randomised one because it is the one that gives independent uniforms.
- `measure` rewritten on top of `_box_partial_cdf`. They agree on every corner set; I left `measure`'s Gray code alone.


lrnv:
On the three not covered points: 

- The SklarDist is a ContinuousMultivariateDistribution, but that's not an issue if the computations are correct, i do not care that much.
- So when marginals are discretes, the rosenblatt algorithm is randomized in the holes, that's what you implemented ? Whazt is the standard in other implementations around (R, python, etc..) ? 
- measure and _box_partial_cdf are basically doing the same operation ? If so, then maybe you could benchmark them and pick the faster, I recall that we engeneered measure() for efficiency a lot, so it might be faster than your version. 

On the rest, give me a bit of time to read the diff.

dcelisgarza:
On the three points:

**Supertype.** Agreed, left alone.

**Rosenblatt on atoms.** Yes: for a discrete coordinate the forward transform draws `V ~ U(0,1)` and returns `H(F(x⁻)) + V·(H(F(x)) − H(F(x⁻)))`, with every later coordinate conditioned on the atom's interval `(F(x⁻), F(x)]`, not on the randomised value; the inverse is deterministic. Randomising within the interval is the *distributional transform* (Rüschendorf 2009, *On the distributional transform, Sklar's theorem, and the empirical copula process*, JSPI 139(11), 3921–3927), and drawing `V` independently to get independent uniforms back is Brockwell (2007, *Universal residuals: A multivariate transformation*, Statistics & Probability Letters 77(14), 1473–1478). Conditioning the successors on the interval is how the discrete pair-copula constructions carry a discrete conditioning variable (Panagiotelis, Czado & Joe 2012, *Pair copula constructions for multivariate discrete data*, JASA 107(499), 1063–1072; Schallhorn, Kraus, Nagler & Czado 2017, *D-vine quantile regression with discrete variables*, arXiv:1705.08310), and the mass itself is Genest & Nešlehová (2007, *A primer on copulas for count data*, ASTIN Bulletin 37(2), 475–515).

This is what vinecopulib does, hence `rvinecopulib` and `pyvinecopulib`: `Vinecop::rosenblatt_impl` in `include/vinecopulib/vinecop/implementation/class.ipp` fills "the second half of U with left-sided limits of the conditional CDF" for the discrete variables and then "randomize[s] by weighting left and right limits with independent uniforms"; its doc comment cites Brockwell (`10.1016/j.spl.2007.02.008`) and says it "is used by default", with `randomize_discrete = FALSE` giving the deterministic upper endpoint. The h-functions it conditions through take `F(x⁻)` alongside `F(x)` for a discrete argument (`AbstractBicop::hfunc1`/`hfunc2` and the `pdf_c_d`/`pdf_d_d` kernels in `bicop/implementation/abstract.ipp`, which are the same finite difference over `[F(x⁻), F(x)]` as the primitive here; their *Discrete variables* docs page describes the layout). R's `copula::cCopula` and statsmodels' copula module are continuous-only and map an atom to `F(x)`, which is what `main` did (the 0.17 row). The alternative I measured — randomise the atom, then condition the successors point-wise on the randomised value — is the obvious shortcut and it is wrong (0.024 against 0.0017), because the randomised latent is independent of everything whereas `U_i` on the atom's interval is not; the test pins that ordering. The five papers are now in `references.bib` and cited from the `condition`, `rosenblatt`, `inverse_rosenblatt` and `SklarDist` docstrings and the manual.

I did some measurements by getting the clanker to make me some scenarios and write out a little table.

**`measure` vs `_box_partial_cdf`.** Same corners when no coordinate is a point, and `measure` was faster; median times, single-threaded:

| case | `measure` | primitive before |
|---|---|---|
| Clayton d=2 | 0.46 µs, 16 allocs | 0.52 µs, 16 |
| Clayton d=3 | 0.49 µs, 12 | 0.67 µs, 24 |
| Gumbel d=3 | 1.21 µs, 32 | 1.37 µs, 44 |
| Gaussian d=3 | 1669 µs | 1656 µs (all `MvNormalCDF`) |
| Clayton d=6 | 5.13 µs, 68 | 6.86 µs, 192 |

The 10–35 % was my `_assemble` allocating a corner vector per term. The `ps == ()` case — the interval conditional's CDF with no point coordinate, and the all-discrete Sklar mass — now dispatches to `measure` and the two are identical on the benchmark; the mixed-partial case has no `measure` equivalent and keeps the corner loop. Two small things in `measure` came with it: its `promote_type(…, Float64)` floor is gone, so `Float32` bounds stay `Float32` through the whole box path (assuming nothing passes Int bounds, I can add the `float` back if it's a problem), and mixed-type corners no longer `InexactError` on the assignment into the `Int` corner vector (`measure(C, (0.0, 0.0, 0.0), (1, 1, 0.5))` errored on `main`). Both are tested.

To keep it that way, `benchmark/benchmarks.jl` gains a `SUITE["measure"]` group (`measure` on 1,000 random 2-D and 3-D boxes and 200 5-D ones, plus the primitive's volume form and its mixed-partial form on the same boxes) so the benchmark tracker sees a regression on either side, and `test/operations/measure.jl` pins the deterministic part: `measure`'s allocation bytes per call, and that the volume form of the primitive allocates exactly the same bytes as `measure` (with a loose 3× minimum-time ratio as a smoke test, since wall-clock in the test suite would only flake on CI).

Full suite after the follow-up: 16,256 passed, 0 failed, 0 errored.

I gave it a clanker-based correctness pass (as I always do now because it keeps catching my mistakes) and it came out ok.

lrnv:
Understood for the three points. 

Can you add all those references to either the documentation and/or the docstrings, anywhere you think they might help a curious reader (look around to find the right syntax) ?

dcelisgarza:
They're already there.

lrnv:
Can you add one or two representative usecases to the benchmark suite, to avoid future regressions ?

dcelisgarza:
I assume you mean of the benchmarking?

lrnv:
Yes, the tachometer benchmarks

dcelisgarza:
I'll add them all, but they might be flaky and will probably have to be updated as github updgrades their runners/julia changes versions.

lrnv:
> I'll add them all, but they might be flaky and will probably have to be updated as github updgrades their runners/julia changes versions.

Why is that ? I found the current ones quite reliable

dcelisgarza:
I think that's more coloured by my experience dealing with the *many* flaky tests and numeric drift over in PortfolioOptimiers' history lol. But there's way more sources of numeric drift over there. Shit breaks every time optim or highs update.

dcelisgarza:
I've updated the comment with the table to mention the benchmarks.

PR #532, opened 2026-09-16
Take observation weights on `fit` and `pseudos`

### LINKED ISSUE #516: Add observation weights to fit.
It's me again, I've got another of those cheeky asks, but I also come with a PR (giving it its clanker pass atm). I accidentally named the issue this first line via the gh CLI.

`weights =` on `fit`: a weighted pseudo-likelihood in the MLE drivers, a weighted `nobs` in `CopulaModel`, and weighted ranks in `pseudos`

I need per-observation weights (exponential decay, importance weights, survey-style reweighting) for a downstream estimator, and I would like to propose the form and offer the PR. I have just done the vine side of this in VineCopulas.jl (Santymax98/VineCopulas.jl#41): the tree criterion, the selection score and the ten pair fitters that package owns all take `weights=`, but two things there stop at the Copulas.jl boundary, and both are the reason for this issue:

- a family fitted by `Distributions.fit(CT, U; method=:mle)` cannot be weighted, because the loss is built inside `_fit`; and
- the `CopulaModel` that `fit(CopulaModel, …)` returns reports the unweighted `loglikelihood` and `nobs` of a weighted fit, so `aic`/`bic` are the unweighted ones.

## Where a weight enters

All references are to `main` at 431b4b2f.

**1. The MLE loss.** The generic transformed-space driver `_fit(CT, U, ::Val{d}, ::Val{:mle})` at `src/Fitting.jl:205` has `loss(C) = -Distributions.loglikelihood(C, U)`. Weighted is `-∑ᵢ wᵢ logpdf(C, U[:, i])`, and `logpdf(C, U::AbstractMatrix)` (`src/Copula.jl:153`) already returns one value per column. The family-specific `:mle` methods build the same quantity by hand and each need the same one-line change:

- `src/ArchimedeanCopula.jl:422`, `f(θ) = -Distributions.loglikelihood(cop(θ), U)` (IPNewton on the one-parameter generators);
- `src/EllipticalCopulas/GaussianCopula.jl:165`, the cross-product `Q = Z * Z'` of the normal scores becomes `Z * Diagonal(w) * Z'` and the `n / 2` in the bivariate objective becomes `∑w / 2`; the `d > 2` branch takes the same weighted `Q`;
- `src/EllipticalCopulas/TCopula.jl:305`, the profile in `λ = 1/ν` calls `_fit(GaussianCopula, U, Val(:mle))` for the endpoint and `_fit_t_corr_given_nu` (`:264`) for each `ν`; the latter builds its own objective, `(n/2) logdet R + (ν + d)/2 ∑ᵢ log1p(qᵢ/ν)`, which weighted is `(∑w/2) logdet R + (ν + d)/2 ∑ᵢ wᵢ log1p(qᵢ/ν)`, and `_t_copula_loglik_factor` takes the same weight;
- `src/ExtremeValueCopula.jl:373` and `src/MiscellaneousCopulas/FGMCopula.jl:171,207`: each has its own `-loglikelihood(…, U)`;
- `src/ArchimaxCopula.jl:182` forwards to the extreme-value driver with `kwargs...`, and the rotation forwarders at `src/MiscellaneousCopulas/SurvivalCopula.jl:232–266` pass `kwargs...` through already, so neither needs anything once the base takes `weights`.

`MCopula`, `WCopula` and `IndependentCopula` have no parameters and ignore it.

**2. The fitted model.** `_run_copula_estimator` at `src/Fitting.jl:408` stores `ll = Distributions.loglikelihood(C, fit_data)`, and `StatsBase.nobs(M::CopulaModel) = size(M.data, 2)` at `:526`. With weights the stored log-likelihood is the weighted one and `nobs` is `∑w`, so `aic` (`:746`), `bic` (`:758`) and the `log(log(n))` criterion (`:766`) follow without change. `nullloglikelihood` (`:779`) evaluates the independence copula on `data`, which is `0` for a copula and needs no weight, but the Sklar path at `:501` (`loglikelihood(S, X)` on the raw data) does.

**3. The pseudo-observations.** `pseudos` (`src/utils.jl:193`) ranks each row unweighted. A weighted rank is the weighted empirical cdf, `Fₙ(xᵢ) = ∑ⱼ wⱼ 𝟙(xⱼ ≤ xᵢ) / ∑ⱼ wⱼ`, with the same tie policy on the weight mass (an `:average` tie block of weight `W_tie` after weight `W_less` takes `W_less + (W_tie + 1)/2` in place of `(i + j)/2`, which is the mean rank of the block in the sample where row `i` is repeated `wᵢ` times). This is the piece VineCopulas.jl deliberately does not do, because it takes `U` as given.

## What I am not proposing here

- **The rank inversions** `:itau`, `:irho`, `:ibeta` (`src/ArchimedeanCopula.jl:352`, `src/ArchimaxCopula.jl:187`, and the pairwise measures behind `_vcov_pairwise_measure` in `src/Inference.jl:34`) go through `StatsBase.corkendall` / `corspearman`, which take no weights. A weighted τ-b is the same O(n log n) Fenwick sweep with weight sums in place of counts; I have one in VineCopulas.jl and can port it, but it is a separate change and I would rather land the likelihood first. Until then `weights` with a rank method should throw, not be ignored.
- **Inference.** `_vcov_hessian` (`src/Inference.jl:44`) differentiates `loglikelihood(…, U)`, so the weighted Hessian is the same one-line change, and the Godambe sandwich (`:82`) weights the per-observation scores; but the sandwich's variance under weights depends on what the weights *are* (frequency weights and importance weights give different sandwiches), which is a design question I would not settle in this PR. `infer` on a weighted model should throw until it is.
- **The margins** in `fit(SklarDist, …)` (`src/Fitting.jl:474`): `Distributions.fit_mle(D, x, w)` exists for most families, so passing the weights through `margins_kwargs` is mechanical, but `Distributions.fit` (which is what is called) does not take them for every family.

## Proposed API

```julia
fit(CT, U; method=:mle, weights=nothing)
fit(CopulaModel, CT, U; method=:mle, weights=nothing)
pseudos(X; weights=nothing, ties=:average)
```

`weights::Union{Nothing, AbstractVector{<:Real}}` of length `n`, non-negative, not all zero, and `nothing` keeps today's code path untouched. Semantics stated in the docstring: the weights multiply the log-likelihood contributions (a weighted pseudo-likelihood), `nobs` is `∑w`, and `pseudos` ranks by weighted mass. Whether to normalise so that `∑w = n` is the one choice to make: VineCopulas.jl does, so that a weight is "how many observations this row counts for" and BIC's `log n` is unchanged for uniform weights; the alternative is to store the raw `∑w` as `nobs` and let the user pick the scale. I lean to normalising, for the same reason.

## Implementation sketch

A single helper,

```julia
_weighted_loglikelihood(C, U, ::Nothing) = Distributions.loglikelihood(C, U)
_weighted_loglikelihood(C, U, w) = sum(j -> w[j] * Distributions.logpdf(C, view(U, :, j)), axes(U, 2))
```

replaces every `-Distributions.loglikelihood(…, U)` in a `:mle` driver. The weighted sum runs over the same column views in the same reduction order as `loglikelihood` (`sum(Base.Fix1(logpdf, d), eachvariate(U))`), so `w ≡ 1` reproduces the unweighted fit bit for bit; a `dot(w, logpdf(C, U))` does not, and I would rather the test be exact. `_run_copula_estimator` threads `weights` into `_fit` through `kwargs...` as it does today, and every `_fit(…, ::Val{:mle})` that ignores it should refuse it rather than drop it, which is a keyword on each method with a `weights === nothing || throw(ArgumentError(…))` where the family is not covered yet.

## Tests

- `fit(CT, U; weights=ones(n))` is bit-identical to `fit(CT, U)` for every family with a `:mle` method.
- Replication identity: a sample in which column `i` is repeated `mᵢ` times, fitted unweighted, equals the deduplicated sample fitted with `weights=m`, to optimiser tolerance; the same for `pseudos` exactly.
- `nobs`, `aic`, `bic` on the weighted `CopulaModel` equal the values on the replicated sample.
- A zero weight removes the row.
- Refusals: negative, non-finite, wrong length, all zero, and `weights` with a rank method.

## Cost

One multiply per column per objective evaluation on the weighted path, nothing on the `nothing` path.

I can send this as a PR in that order: the likelihood helper and the generic driver, then the family-specific drivers, then `CopulaModel`, then `pseudos`.

dcelisgarza:
Clanker audit done, I'm not sure this is out of the API but let me know if I should change anything.

Closes #516.

## What

`fit(CT, U; weights)`, `fit(CopulaModel, CT, U; weights)`, both `NestedArchimedeanCopula` fit forms, and `pseudos(X; weights)` take one non-negative, finite weight per observation, not all zero. `nothing` keeps today's code path untouched.

Semantics, stated once in the `fit(CopulaModel, …)` docstring and in the manual:

- the weights are normalized to sum to `n`, so a weight is "how many observations this column counts for", the estimator is invariant to their scale, uniform weights reproduce the unweighted fit **bit for bit**, and integer weights summing to `n` give the fit of the sample in which each observation is repeated that many times;
- the likelihood drivers maximize `∑ⱼ wⱼ log c(uⱼ)`; a zero weight removes its observation, even where the density vanishes;
- `CopulaModel` stores the weighted log-likelihood, so `aic`, `bic`, `deviance` and the selection criteria follow; `nobs` stays `n`; the recipe carries the normalized weights and `show` says so;
- with `pseudo_values=false` the ranks are the weighted ones of `pseudos(X; weights)`: observation `i` takes the mean rank of its copies in the replicated sample under every tie convention, divided by `n + 1`; a zero-weight tie group sits at the weighted ECDF of its value so every rank stays strictly inside `(0, 1)`.

Refusals: a rank-inversion method, the Sklar route (its margins are fitted unweighted), and `infer` / `_refit` / composite `GOFCopulaTest` on a weighted model.

## How

One helper carries all of it:

```julia
_weighted_loglikelihood(D, X, ::Nothing) = Distributions.loglikelihood(D, X)
_weighted_loglikelihood(D, X, w) = sum(j -> w[j] * logpdf(D, view(X, :, j)), axes(X, 2))  # zero-weight guard inside
```

The weighted sum runs over the same column views in the same order as `Distributions.loglikelihood`, which is what makes `w ≡ 1` bit-identical. The generic transformed-space driver, the Archimedean IPNewton driver, the extreme-value, FGM and nested Archimedean objectives call it. Two families have hand-written objectives:

- **Gaussian**: `Q = Z * Z'` becomes `Zw = Z .* sqrt.(w)'; Q = Zw * Zw'`, the root scaling keeps the product a symmetric rank-k update on the same BLAS kernel (`(Z .* w') * Z'` goes through `gemm` and is not bit-identical for `w ≡ 1`); `n` becomes `∑w`.
- **Student**: `_t_copula_loglik_factor` and `_fit_t_corr_given_nu` weight their per-observation sums and use `∑w` in front of the constants; the profile over `λ = 1/ν` forwards `weights` to the Gaussian endpoint and to every inner fit.

`_run_copula_estimator` validates and normalizes once (`_fit_weights`), refuses non-likelihood methods, ranks by the already-normalized weights on the `:mpl` path, and forwards `weights` to `_fit`. The parameter-free families take and ignore the keyword. A family whose own `_fit(…, ::Val{:mle})` has no keyword arguments is fitted by the generic driver whenever any keyword is passed, that is how keyword dispatch already behaves today, and the developer guide now says so.

## Not in this PR (as in #516)

- The rank inversions `:itau`, `:irho`, `:ibeta` (`corkendall` / `corspearman` take no weights), they throw.
- Inference on a weighted model throws; the sandwich depends on what the weights are.
- Weighted margins in `fit(SklarDist, …)`, the Sklar route throws with a pointer to `pseudos(X; weights)` + `fit(CT, U; weights)`.

## Tests

- Unit weights at three scales (`ones`, `fill(3, n)`, `fill(2.5, n)`) are bit-identical to the unweighted `CopulaModel` (coefficients, log-likelihood, `aic`, `bic`) in fifteen families: Clayton, Gumbel (d=3), Frank, AMH, BB1, Gaussian (d=2, 3), Student (d=2, 3), FGM (d=2, 3), Galambos, a survival and a rotated copula, and the independence copula.
- Integer weights summing to `n` reproduce the fit of the replicated sample to optimizer tolerance (Brent's default `rel_tol = √eps` caps it around `1e-7`), the log-likelihood and BIC to `1e-8`, and the estimator is scale invariant.
- A zero weight removes its observation, including a column at `1e-300` where the density vanishes.
- The MPL path equals a fit on `pseudos(X; weights)`; `_copula_data`, `nullloglikelihood` and `residuals` agree.
- Selection over candidates and both nested Archimedean forms (template and custom `reparam`) take weights.
- `pseudos`: every tie convention is bit-identical for unit weights; integer counts reproduce the replicated sample's mean ranks under every convention; the zero-weight cases are pinned by hand; every refusal.

Locally (Julia 1.13, single thread) all pass: `Aqua.jl`, `api/utilities.jl` (127), `operations/fitting.jl` (1697, of which the new `observation weights` testset is 311), `operations/inference.jl` (23), `operations/fitting_selection.jl` (41), `operations/hypothesis_testing.jl` (173), `correctness/elliptical.jl` (30), `correctness/nested_archimedean.jl` (85), `correctness/extreme_value.jl` (364), `correctness/bivariate_families.jl` (60), `correctness/archimedean.jl` (36).

Side note found while writing the tests: `show(::CopulaModel)` on a `d = 3` Gaussian model takes minutes because `_print_dependence_metrics` computes Spearman's ρ by `hcubature` over `MvNormalCDF`. Pre-existing, not touched here.


dcelisgarza:
The benchmark failure appears to be bad luck as that's a file this PR doesn't touch.

dcelisgarza:
I'm writing a stacked PR to address the Not in this PR section, as that's a complete solution.

lrnv:
@dcelisgarza I was going to say that for SklarDist, you should simply call the marginal weighted fits and let them fail graciously, but now i'll wait for your take on the stacked PR. First time i see stacked PRs, you'll have to drive me through it 🤣

dcelisgarza:
First time i'm using them too. They've just gotten supported by github. We'll learn together lol.

dcelisgarza:
Ok so apparently a stack of PRs just wait for the ones in the bottom to be resolved, then those changes get propagated up via rebases or resolved as merge conflicts.

lrnv:
So this one is frist in line, let me check the code once more and then i'll merge it

dcelisgarza:
I've been giving them clanker based audits as well. I know this is your baby and I don't want to ruin it, so please no rush.

lrnv:
@dcelisgarza here's a few notes on the implementation: 

### 1. Zero-weights do not disapear in Gaussian and Students.

The documented contract says that a zero weight removes an observation, but it looks like on the boudary of the hypercube (values of 0 or 1) with a zero weight will produce a NaN output. I would simply exclude zero-weights by filtering the data, would be cleaner and would match the "zero weight removes the observation" semantics

### 2. Weighted `pseudos` should promote with the weight type

The weighted `pseudos` implementation currently chooses

```julia
T = float(eltype(sample))
```

for its working/output type.

I think this should include the weight type as well, e.g. along the lines of

```julia
T = float(promote_type(eltype(sample), eltype(weights)))
```

Otherwise a `Float64` sample with `BigFloat` weights silently performs the weighted rank arithmetic in `Float64`.

This seems especially worth preserving now that the rest of the package is increasingly generic over `Real` inputs.

### 3. Rebase carefully and follow optimizer changes.

Please rebase on main, and take into account that we remove the `try LBFGS / catch NelderMead` optimizer setup in favor of a graciously failling LBFGS, so please do the same. 

I'll look at the next PR once this one is merged.

dcelisgarza:
Bloody clanker pass added the type conversion and i didn't realise. Working on the fixes now.

dcelisgarza:
Thanks, all three addressed in the last commit and the branch is rebased on `main`.

**1. Zero weights.** One helper, `_weighted_sample(X, weights)`, drops the zero-weight columns and their weights, and every likelihood engine is handed that pair: `_run_copula_estimator` for the families and `_fit_nested` for the nested Archimedean fits. The `ifelse` that special-cased a zero term in `_weighted_loglikelihood` is gone with it, so the engines never meet a zero weight and the Gaussian/Student cross-products never see `0 * Inf`. The model still stores every column and its normalized weight, so `nobs`, `_copula_data` and the recorded recipe are as before. New tests fit `GaussianCopula`, `TCopula` and `ClaytonCopula` with a zero-weight column sitting at `0` and at `1` and compare against the fit without that column.

**2. `pseudos` working type.** Now `float(promote_type(eltype(sample), eltype(weights)))`, as you suggested. Tested for `Float32`/`Float32 → Float32`, `Float32`/`Float64 → Float64`, and `Float64`/`BigFloat → BigFloat` (whose `Float64` rounding equals the plain result). One existing assertion changed with it: `pseudos(Float32.(X); weights=ones(n))` is now `Float64` since `ones(n)` is `Float64`; `ones(Float32, n)` keeps `Float32`.

**3. Rebase.** Rebased on `main` at b7dbd20. The weighted objectives follow #545: a single LBFGS call with no `try`/`catch NelderMead` fallback, in the generic driver, the extreme-value, Student and nested Archimedean fits.

Ran `api/utilities.jl`, `operations/fitting.jl`, `correctness/elliptical.jl` and `correctness/nested_archimedean.jl` locally; all green.

dcelisgarza:
I'm updating the stacked pr to incorporate the fixes and rebase.

dcelisgarza:
Looks like we waded into some very similar numerical issues with Optim i've had over at PortfolioOptimisers. I'm getting the clanker to solve it.

dcelisgarza:
The two red jobs are one test, `hypothesis_testing.jl:438` (`Composite GOF preserves runtime fitting structure`), which raises `DomainError` from `composition_taylor` inside the nested template fit. It is not the weights: the same fit through `_fit_nested(recon, α₀, U)` and through `_fit_nested(recon, α₀, U; weights=nothing)` takes two LBFGS trajectories on this branch, and `main` clears the test on the first one only. The underlying defect is on `main`: `_nested_rebound` skips the nesting certificate, so LBFGS can propose a child Clayton θ below its parent's, or below zero, where the composed generator is not defined; #545 removed the Nelder–Mead fallback that used to mask the `DomainError`. A bootstrap of a Clayton template hits it on three of four seeds on `main` too.

The last commit keeps the fit in the certified region: the loss of `_fit_nested` is `Inf` on a tree that fails a certificate, as the Student objective is on a singular correlation factor, and every constructible template starts inside. `hypothesis_testing.jl` (173), `nested_archimedean.jl` (85) and `fitting.jl` pass locally on 1.13 with it; the bootstrap test covers the four seeds. If you would rather take that as its own PR, or fix it in the coordinates (a child written as `θ_parent + exp(α)`), say so and I will split it out.

dcelisgarza:
Since the clanker found #550 and #551 ill get it to fix them.

dcelisgarza:
The Aqua failure is from an Aqua update. I'm getting the same failure myself.

dcelisgarza:
The two red jobs are one test, `hypothesis_testing.jl:438` (`Composite GOF preserves runtime fitting structure`), which raises `DomainError` from `composition_taylor` inside the nested template fit. It is not the weights: the same fit through `_fit_nested(recon, α₀, U)` and through `_fit_nested(recon, α₀, U; weights=nothing)` takes two LBFGS trajectories on this branch, and `main` clears the test on the first one only. The underlying defect is on `main`: `_nested_rebound` skips the nesting certificate, so LBFGS can propose a child Clayton θ below its parent's, or below zero, where the composed generator is not defined; #545 removed the Nelder–Mead fallback that used to mask the `DomainError`. A bootstrap of a Clayton template hits it on three of four seeds on `main` too.

The last commit keeps the fit in the certified region: the loss of `_fit_nested` is `Inf` on a tree that fails a certificate, as the Student objective is on a singular correlation factor, and every constructible template starts inside. `hypothesis_testing.jl` (173), `nested_archimedean.jl` (85) and `fitting.jl` pass locally on 1.13 with it; the bootstrap test covers the four seeds. If you would rather take that as its own PR, or fix it in the coordinates (a child written as `θ_parent + exp(α)`), say so and I will split it out.

lrnv:
@dcelisgarza i'm confused, do you want me to take a look or do i wait for CI to be green ? You can put it in draft mode untill you think its ready for review

dcelisgarza:
Two red checks, two causes.

**Aqua.** The `Persistent tasks` error (`Unable to locate ChainRulesCore, a dependency of SpecialFunctions`) is Aqua 0.8.17, released 16:46 UTC today, minutes before this run; `main`'s last green run resolved 0.8.16. Aqua 0.8.18 (its #400, "no longer errors on a weak dependency that is also listed in `[deps]` but not installed") is in General already, so the next run resolves it without a change here. Whether to pin `Aqua` in the test compat is yours; I have not touched it.

**Benchmark.** `fitting/gaussian_mle` +6% was this branch: the weighted Gaussian fit reassigned `n` (`d, n = size(Udata)` then `n = weights === nothing ? n : sum(weights)`), and `n` is captured by both objectives, so it became a `Core.Box` and `n / 2 * log(…)` dispatched dynamically on every evaluation. The last commit assigns it once; locally `_fit` is back to `main`'s 376 allocations and 74 KiB from 428 and 77 KiB. The fitting tests (345 in `observation weights`) pass with it.

PR #534, opened 2026-09-16
Propagate inverse-Williamson CDF NaNs

### LINKED ISSUE #523: Generic inverse-Williamson CDF silently converts NaN to probability one
## Problem

The generic inverse-Williamson distribution currently ends its CDF evaluation with:

```julia
F = 1 - rez
# Guard against tiny numerical excursions
return isnan(F) ? one(x) : clamp(F, zero(x), one(x))
```

Clamping a small **finite** excursion is reasonable. Converting `NaN` to `1` is not: `NaN` can indicate a failed derivative evaluation, an invalid downstream `Generator` implementation, overflow/cancellation, or another numerical bug. Returning certainty at that point silently turns a failure into a valid-looking CDF value.

This is especially risky because `𝒲₋₁` is public and the generic inverse distribution is used by Archimedean sampling. A spurious `F(x)=1` can then affect quantile inversion and random generation rather than surfacing the underlying problem.

The policy is also inconsistent with the shared quantile engine, which explicitly rejects a NaN objective with

```julia
throw(ArgumentError("the CDF returned NaN at x = $x"))
```

## Suggested behavior

- Clamp only finite numerical excursions outside `[0,1]`.
- Preserve/raise a diagnostic for `NaN` rather than converting it to one. Whether direct `cdf` should return `NaN` or throw can be decided consistently with the rest of the distribution API; the important part is not to manufacture a probability.
- If there is a specific mathematically justified endpoint at which an intermediate expression is `NaN` but the limit is exactly one, handle that endpoint explicitly before the generic computation instead of treating every NaN as that case.

## Tests

- A deliberately incomplete/pathological test generator whose derivative produces NaN must not yield `cdf(𝒲₋₁(...), x) == 1` at an interior point.
- Existing valid generators still have CDF values in `[0,1]` and reach one at their true upper limit.
- Quantile inversion propagates the diagnostic rather than silently accepting the NaN as a bracket endpoint.
- Boundary/overflow regression tests distinguish finite roundoff from NaN.

lrnv:
Closes #523.

This PR makes the generic inverse-Williamson CDF propagate `NaN` instead of converting it to probability one, while treating the mathematically known `+Inf` endpoint explicitly. Quantile inversion already rejects a NaN CDF value, so sampling will surface the failure rather than silently accepting it.

Regression tests are being added on this branch before review.


lrnv:
LGTM

PR #537, opened 2026-09-16
Scope README compatibility claims for 1.0

### LINKED ISSUE #531: Follow-up to #499: README still overclaims full Distributions and numeric/GPU compatibility
## Problem

#499 explicitly identified two homepage/README claims as too broad for the 1.0 contract, but both are still present on current `main` after that issue was closed.

The README still describes the package as:

> A fully `Distributions.jl`-compliant copula package

and says:

> This compliance allows direct interoperability ...

It also still makes the package-wide claim:

> a notable feature ... is floating point type agnosticism, i.e., compatibility with `Float32` (and GPU equivalents), `BigFloat`, `DoubleFloats`, `MultiFloats`, and other numeric types.

The latter is demonstrably stronger than the current implementation: structured EV constructors still canonicalize to Float64 (#524), several conditioning specializations are Float64-only (#530), inference/testing contains explicit Float64 work arrays, and GPU compatibility is not established as a package-wide contract.

The public API page is now much more carefully scoped, so the README currently advertises a broader contract than the canonical API documentation.

## Suggested fix

Apply the wording already proposed in #499 rather than creating a new promise:

- replace absolute “fully compliant” language with something like “integrates with the Distributions.jl distribution API”;
- replace package-wide numeric/GPU compatibility with a scoped statement such as “core numerical paths are generic where practical; individual models/backends may document narrower numeric support”;
- keep the Public API page as the normative compatibility contract.

Because `docs/src/index.md` is synchronized from the README, verify both rendered entry points after the change.

## Acceptance criteria

- README/index no longer claim blanket Distributions.jl compliance beyond the documented generalized semantics for singular/mixed objects.
- README/index no longer promise GPU, BigFloat, DoubleFloats, MultiFloats, etc. package-wide unless CI actually establishes that support.
- The homepage links readers to the Public API page for the precise 1.0 guarantees.
- The wording remains accurate if #524/#530 are resolved only partially rather than by full arbitrary-precision support.

This is a follow-up to the still-valid documentation findings in #499, not a new API policy proposal.

lrnv:
Closes #531.

## What

- replaces the absolute “fully Distributions.jl-compliant” homepage tagline with integration wording;
- scopes numeric compatibility to the documented model/backend contract instead of promising package-wide Float32/GPU/BigFloat/DoubleFloats/MultiFloats support;
- links the README to the Public API page as the normative 1.0 contract;
- updates the generated-homepage frontmatter tagline. `docs/make.jl` runs `docs/sync_homepage.jl`, so the rendered `docs/src/index.md` is regenerated from this README during documentation builds.

The historical JOSS paper title in the citation block is intentionally unchanged.


lrnv:
LGTM

PR #539, opened 2026-09-16
Define copula entropy through KL measure semantics

### LINKED ISSUE #527: Copula entropy fallback uses generalized densities on singular copulas
## Bug / API mismatch

The public `ι(C)` docstring defines copula entropy as the expected negative **Lebesgue log-density** and explicitly says that it is not a definition for singular copulas:

```julia
ι(C::Copula) = expectation(u -> -logpdf(C, u), C; nsamples=10^4)
```

However, the generic method does not inspect `copula_measure_style(C)`. Copulas.jl deliberately gives singular models generalized `pdf`/`logpdf` semantics for other public operations, so feeding that generalized value into the differential-entropy formula changes the mathematical quantity.

A minimal example is `WCopula()`: samples lie on the anti-diagonal and its generalized `_logpdf` is `0` on that support, so the generic entropy route can report `ι(WCopula()) == 0`, indistinguishable from independence, even though the singular copula has no ordinary copula density and the documented entropy is not defined by this formula.

The same risk applies to other `NonAbsolutelyContinuousMeasure()` models (singular limits, Raftery, discrete-spectral EV copulas, some Liouville/Archimax constructions, etc.) unless they each happen to override `ι`.

## Suggested contract

Gate the generic differential-entropy fallback on `copula_measure_style(C) === AbsolutelyContinuousMeasure()`.

For non-absolutely-continuous copulas, choose one explicit public behavior before 1.0:

- throw an informative error that ordinary copula entropy is unavailable; or
- define a separate/documented generalized measure if the package really wants one.

Do not silently reuse generalized-density semantics inside a formula whose docstring requires a Lebesgue density.

## Tests

- `ι(IndependentCopula(2)) == 0` remains valid.
- A regular dependent continuous copula returns its current finite estimate.
- `WCopula`, `MCopula`, and at least one nontrivial singular/mixed copula do not silently pass through the continuous entropy fallback.
- Survival/subset wrappers preserve the measure-style gate.
- The behavior agrees with the public dependence-measures documentation.

lrnv:
Closes #527.

## What

- defines public copula entropy as `-D_KL(P_C || λ^d)`;
- routes the generic implementation through the existing `copula_measure_style` trait;
- keeps `E[-logpdf]` only for absolutely continuous copulas;
- returns `-Inf` for any non-absolutely-continuous copula, including singular/mixed laws and `EmpiricalCopula`;
- documents that atomic Shannon entropy is a different quantity;
- adds regression coverage for independence, W, Raftery, empirical, and a regular continuous copula.

No generalized singular `logpdf` value is reused in the entropy definition.


lrnv:
LGTM

PR #540, opened 2026-09-16
Validate Sklar fitting-target margin arity

### LINKED ISSUE #529: SklarDist fitting target does not validate the number of margin families
## Bug

`SklarDist{CopulaType,Tuple{MarginTypes...}}` is explicitly documented as a supported public fitting-target syntax for 1.0, but `_estimate_sklar` does not validate that the number of marginal families equals the data dimension.

The fitter currently does:

```julia
d = size(X, 1)
m = ntuple(i -> Distributions.fit(TplMargins.parameters[i], @view X[i, :]; ...), d)
```

This produces two different bad public behaviors:

1. **Too few margin families**: indexing `TplMargins.parameters[i]` eventually raises an incidental `BoundsError`.
2. **Too many margin families**: `ntuple(..., d)` simply never reads the extras, so the additional public family selectors are silently ignored and a different fitting target is used than the type the caller wrote.

For example, with two rows of data, a target containing three margin types can fit only the first two without warning.

## Why this is 1.0-critical

The canonical public-API page singles this exact parameterized type syntax out as a stable exception to the rule that concrete type parameters are implementation details. Its arity therefore needs an explicit validation contract rather than relying on tuple indexing.

## Suggested fix

Validate before fitting any margin:

```julia
length(TplMargins.parameters) == d || throw(DimensionMismatch(...))
```

(or the equivalent robust type-level query), with an error message identifying both the data dimension and the number of requested margin families.

It would also be useful to validate that every tuple entry is an admissible univariate distribution family before beginning partial work, so malformed public targets fail atomically and informatively.

## Tests

- exact-length heterogeneous margin tuples fit normally;
- too few margin families throw `DimensionMismatch`, not `BoundsError`;
- too many margin families throw instead of silently ignoring extras;
- both `fit(SklarDist{...}, X)` and `fit(CopulaModel, SklarDist{...}, X)` share the behavior;
- refitting/bootstrap replay preserves the same validated target arity.

lrnv:
Closes #529.

## What

- validates the public `SklarDist{CopulaType,Tuple{MarginTypes...}}` fitting target before any marginal fit starts;
- requires exactly one margin family per data row and throws `DimensionMismatch` otherwise;
- validates that every margin selector is a univariate distribution family;
- applies the same validation to `fit(SklarDist{...}, X)` and `fit(CopulaModel, SklarDist{...}, X)`;
- adds focused regression tests for exact, too-short, too-long, and malformed targets.

The estimator machinery itself is unchanged; validation is isolated in a small fitting-target layer.


lrnv:
LGTM

PR #541, opened 2026-09-16
Preserve numeric types in generic Taylor padding

### LINKED ISSUE #526: Generic Generator derivative fallback narrows padded Taylor coefficients to Float64
## Problem

`Generator` is a documented public extension point, and the public contract says that its generic AD/Taylor fallbacks provide density and sampling behavior when a downstream generator implements only `ϕ`, `max_monotony`, and `params`.

The shared internal `taylor` helper currently has a numeric-type hole when `TaylorSeries.jl` returns fewer than `d + 1` coefficients:

```julia
rez = f(x₀ + TaylorSeries.Taylor1(eltype(x₀), d)).coeffs
...
if p < d+1
    v = zeros(d+1)       # Vector{Float64}
    v[1:p] .= rez
    return v
end
```

The comment immediately above this branch notes that shorter coefficient vectors are an expected TaylorSeries behavior. In that case a `BigFloat`, `Float32`, or AD-compatible coefficient sequence is silently copied into a `Vector{Float64}`.

`ϕ⁽ᵏ⁾(G::Generator, ...)` consumes the last element of this helper, so this can leak directly into the public generic generator fallback. It also makes the result type depend on whether TaylorSeries happened to retain trailing zero coefficients.

## Why this matters for 1.0

The public API explicitly promises numeric representation propagation where practical, and the downstream `Generator` extension contract relies on the generic derivative machinery. A type-changing fallback hidden behind an implementation-detail helper breaks that behavioral contract even though `taylor` itself is private.

## Fix

Allocate padding using the Taylor coefficient element type, e.g. a `similar`/`zeros(eltype(rez), ...)` style construction that also works for nested AD types, rather than unconditionally using `Float64`.

Audit the other generic generator fallback allocations in the same pass for the same pattern.

## Tests

- A downstream-style `Generator` using the generic `ϕ⁽ᵏ⁾` path preserves `Float32` and `BigFloat` when TaylorSeries returns a shortened coefficient vector.
- Zero higher derivatives are returned as `zero(T)`, not `0.0` of another type.
- The generic density / inverse-Williamson path remains type-consistent for the same test generator.
- Existing Float64 behavior is unchanged.

lrnv:
Closes #526.

## What

- adds a scalar-`Real` specialization of the shared Taylor helper used by generic generator derivatives;
- pads shortened Taylor coefficient vectors with `zero(eltype(rez))` instead of `Float64` zeros;
- preserves `Float32`, `BigFloat`, and nested real/AD coefficient types without changing the existing generic fallback for non-real expansion points;
- adds a downstream-style generator fixture proving type preservation through `ϕ⁽ᵏ⁾` and the generic inverse-Williamson CDF/PDF path.

Float64 behavior remains unchanged.


lrnv:
LGTM

Issue #521, opened 2026-09-16
Zero-dimensional copulas are constructible through public constructors


lrnv:
## Problem

Several parameter-free public constructors permit `d = 0`, even though the public API assumes a positive copula dimension and individual docs such as `IndependentCopula` explicitly state `d ≥ 1`.

For example, both of these currently construct an object:

```julia
IndependentCopula(0)
IndependentCopula{0}()
```

The resulting object is not usable with the generic distribution interface. A matching empty input reaches generic `cdf`, where the all-above-one branch is vacuously true and the implementation tries to evaluate `one(u[1])`, producing a `BoundsError`. Similar empty-dimension failures exist for other copulas whose formulas call `minimum`, inspect a first coordinate, or construct lower-dimensional conditional objects.

This is a constructor-invariant bug: the invalid object should never exist.

## Suggested direction

Define a package-wide minimum-dimension rule for `Copula{d}` and enforce it in public dimension-bearing constructors. If dimension one is intentionally supported for some families (as documented for independence), keep that behavior; this issue is specifically about `d <= 0`.

A small internal dimension validator may be preferable to repeating checks in every parameter-free family, but it should not make valid family-specific restrictions less explicit (`WCopula` remains bivariate-only, for example).

## Acceptance criteria

- Public copula constructors reject `d <= 0` with the stable public exception taxonomy.
- `IndependentCopula(1)` / `{1}()` continue to work if the documented `d ≥ 1` contract is retained.
- Dimension-specific families continue to enforce their stronger restrictions.
- Generic public operations cannot receive a zero-dimensional `Copula` produced by ordinary public constructors.
- Regression tests cover runtime-dimension and type-parameter constructor forms.

lrnv:
## 1.0 API decision (from #525)

The minimum public copula dimension has now been decided more strongly than this issue originally proposed:

> **All public copulas must satisfy `d ≥ 2`.**

So this issue should be implemented as a package-wide `d < 2` constructor-invariant check, not merely a `d <= 0` check. `IndependentCopula(1)` is intentionally no longer part of the 1.0 public API.

This is consistent with existing reduction APIs: selecting one coordinate from a copula already returns `Distributions.Uniform()` rather than constructing a `Copula{1}`. The one-dimensional mathematical identity copula therefore remains representable where needed without introducing a dedicated one-dimensional copula object.

Acceptance criteria should therefore be read as:

- all public copula constructors reject `d < 2` with the stable public exception taxonomy;
[...]

PR #543, opened 2026-09-16
Define the dependence-measure dimension contract

### LINKED ISSUE #525: Define the public dependence-measure contract for one-dimensional copulas
## Problem

`IndependentCopula` explicitly supports `d ≥ 1`, so `IndependentCopula(1)` is a valid public object. The generic scalar dependence measures, however, use multivariate normalizations that are singular at `d = 1`:

- Spearman `ρ`: denominator `2^d - d - 1 = 0`;
- Kendall `τ`: denominator `2^(d-1) - 1 = 0`;
- Blomqvist `β`: denominator `2^(d-1) - 1 = 0`;
- Gini `γ`: the independence and comonotonicity normalizers coincide at `d = 1`, so `b(d)-a(d)=0`.

Thus public calls such as `Copulas.ρ(IndependentCopula(1))` currently fall into undefined `0/0`-style arithmetic rather than a deliberate API behavior.

This is separate from rejecting zero-dimensional copulas (#521): dimension one is currently documented as supported and can legitimately arise as a model object.

## 1.0 decision needed

Choose and document one of these contracts:

1. dependence summaries are defined only for `d ≥ 2`, and every public scalar measure rejects a one-dimensional copula with a stable informative exception; or
2. Copulas.jl adopts explicit conventional values for `d = 1`, with those conventions documented and tested.

Silently returning `NaN`/`Inf` from the generic normalization should not be the public 1.0 behavior.

The same decision should be applied consistently to data-based scalar summaries on `1 × n` input where their normalizations are likewise degenerate.

## Acceptance criteria

- `τ`, `ρ`, `β`, and `γ` have intentional, documented behavior for `Copula{1}` and `1 × n` data.
- No one-dimensional call reaches a zero denominator accidentally.
- The public API/dependence-measures documentation states the supported dimensional domain or adopted convention.
- Tests cover `IndependentCopula(1)` and one-dimensional empirical inputs.
- Dimension-two and higher behavior is unchanged.

lrnv:
Closes #525.

## What

Implements the remaining consequences of the 1.0 decision that public copulas have `d ≥ 2`:

- documents the package-wide public copula dimension domain and the `Uniform()` representation of one-coordinate reductions;
- explicitly states the scalar dependence-measure domain in the canonical Public API contract;
- rejects one-row data inputs for multivariate `τ`, `ρ`, `β`, and `γ` with `DimensionMismatch` instead of reaching degenerate normalizations;
- leaves the one-dimensional data entropy estimator `ι(U)` available, because entropy itself is not one of those singular concordance normalizations;
- adds regressions proving 2D+ behavior remains available.

Constructor enforcement itself is tracked independently in #521.


lrnv (commented):
LGTM

lrnv (commented):
LGTM

PR #546, opened 2026-09-16
Finish the 1.0 public exception taxonomy

### LINKED ISSUE #519: Finish the 1.0 public exception taxonomy after #498
## Problem

#498 established the intended public failure taxonomy for 1.0:

- wrong dimension → `DimensionMismatch`;
- parameter outside a mathematical domain → `DomainError`;
- invalid option/index/layout → `ArgumentError`;
- `@assert` reserved for internal invariants.

The recent cleanup fixed many paths, but current `main` still exposes a large number of dimension failures as `ArgumentError`, and at least one public constructor still uses raw `error(...)`.

Representative public paths:

- generic `Copula` vector/matrix `cdf`, `logpdf`, and `_rand!` dimension checks;
- `SklarDist` vector/matrix `cdf`, `pdf`, `logpdf`, and sampling checks;
- `SubsetCopula`, `ArchimedeanCopula`, `LiouvilleCopula`, and many concrete-family matrix samplers;
- `TCopula` Rosenblatt dimension validation;
- `MCopula`, `WCopula`, `BetaCopula`, `RafteryCopula`, `PlackettCopula`, reflected copulas, `BernsteinCopula`, etc.;
- `WCopula{d}()` currently calls `error("WCopula only available in dimension 2")` for an invalid public dimension.

Because these are public calls, the exception type is part of the API contract we just chose to stabilize.

## Suggested direction

Do one final systematic sweep of public argument validation instead of fixing the remaining methods opportunistically. Prefer a shared dimension-check helper only if it improves readability without obscuring the public error messages.

Internal `@assert`s that are genuinely unreachable after public validation can remain assertions.

## Acceptance criteria

- Public vector/matrix shape mismatches consistently throw `DimensionMismatch`.
- Public invalid dimensions/options/indices consistently throw `ArgumentError` unless they are mathematical-domain violations.
- Public mathematical-domain violations use `DomainError` where appropriate.
- No public constructor path uses raw `error(...)` / `ErrorException` for ordinary invalid user input.
- Regression tests cover the generic `Copula`/`SklarDist` interfaces plus representative specialized methods so specialization cannot silently change the exception contract.

This is a follow-up audit item rather than a change of policy: the policy from #498 is good, but the sweep is not yet complete.

lrnv:
Closes #519.

## What

- makes the generic real-valued copula CDF/log-density matrix paths report shape mismatches as `DimensionMismatch`;
- aligns `SklarDist` vector/matrix CDF, density, log-density, and sampling shape checks with the same contract;
- replaces the public `WCopula` raw `ErrorException` constructor failure with `ArgumentError`;
- updates representative singular-family samplers (`MCopula`/`WCopula`) to use `DimensionMismatch` for output-buffer shape errors;
- keeps domain/option failures distinct from shape failures.

## Tests

Adds regressions for generic copula and Sklar vector/matrix calls, specialized singular samplers, and invalid `WCopula` dimensions.


lrnv:
Isnt the content of the public exception contract overwriting methods that are defined in src/Copula.jl and src/SklarDist.jl ? In any case, these methods should be in the src/Copula.jl file IMHO and not in a random file included in another random file. 

I feel like we overwritten methods instead of modifying them simply, which is bad.

lrnv:
LGTM

PR #548, opened 2026-09-16
Take observation weights on the rank inversions, in inference and on the Sklar route

### LINKED ISSUE #547: `weights=` beyond the likelihood: the rank inversions, `infer`, and the Sklar margins
#516 stopped at the likelihood on purpose, and its PR #532 refuses a weight everywhere else. I would like to propose the three pieces its "what I am not proposing" section left out, in one stacked PR, because once a weight has a meaning the other three fall out of it and I would rather not leave three refusals in the code that the docs have to explain. Same status as #516: documented, outside the public contract, however that is settled there.

The meaning #532 gave a weight is "how many observations this column counts for": the weights are normalized to sum to `n`, unit weights reproduce the unweighted fit, and integer weights summing to `n` give the fit of the sample in which observation `j` is repeated `weights[j]` times. Everything below is that reading carried through: each procedure is the unweighted one on the replicated sample, written so that it extends to real weights and is invariant to their scale. Nothing here decides between frequency and importance weights; it says which one `fit` already chose and follows it.

All references are to the tip of #532 (e80abb7c) unless they say `main`.

## 1. The rank inversions

`_run_copula_estimator` refuses `weights` with a rank method at `src/Fitting.jl:476`. The sample measures behind `:itau`, `:irho`, `:ibeta` and the Student `:itau_irho` are `StatsBase.corkendall` / `corspearman` (`src/Fitting.jl:308`, `src/ArchimedeanCopula.jl:359`, `src/EllipticalCopulas/GaussianCopula.jl:235,239`, `src/EllipticalCopulas/TCopula.jl:354,392`, `src/ExtremeValueCopula.jl:334`), `corblomqvist` (same sites), and the multivariate `β(U)` at `src/Copula.jl:339`. None of them takes weights. Their weighted forms are the measures of the replicated sample:

- **Kendall's tau-b.** `τ = Σ_{i<j} wᵢwⱼ sgn(xᵢ−xⱼ) sgn(yᵢ−yⱼ) / √((W₀−W₁)(W₀−W₂))`, with `W₀ = Σ_{i<j} wᵢwⱼ` and `W₁`, `W₂` the weighted tie pairs `Σ_g (T_g² − Σ_{i∈g} wᵢ²)/2` over the tie groups of `x` and `y`. The pairs formed by the copies of one observation cancel between numerator and denominator, so for integer weights this is exactly the tau-b of the replicated sample, and `W₀ − W₁ = (W² − Σ_g T_g²)/2` is what makes it so. It is one O(n log n) sweep over `x` with a Fenwick tree over the dense ranks of `y`, each `x`-tie block queried before it is inserted, weight sums in place of counts. Its arithmetic reproduces `corkendall`'s (`(exact integer) / sqrt(float(a) * float(b))`) so unit weights are bit-identical.
- **Spearman's rho.** The weighted Pearson correlation of the weighted average ranks (a tie block of weight `T` after weight `B` takes rank `B + (T+1)/2`, which is `tiedrank` for unit weights). Exact for the replicated sample; within an ulp of `corspearman` for unit weights, since it is a different reduction.
- **Blomqvist's beta.** The same weighted average ranks split at the median rank `(W+1)/2` of the replicated sample, concordant mass counted; bit-identical for unit weights. The multivariate `β(U)` weights its orthant count.

One selector, `_rank_measure(::Val{:itau}, U, weights)`, returns `corkendall(U')` for `nothing` and the weighted kernel otherwise, and each rank `_fit` gains `weights=nothing` and calls it; the parameter-free families take and ignore the keyword as they do for `:mle`. `:iupper` and the nonparametric estimators keep refusing, `_cortail` is its own machinery.

## 2. Inference

`infer` refuses a weighted model at `src/Inference.jl:305` and `_refit` at `src/Fitting.jl:124`. Under the reading above every procedure is defined:

- **`:hessian`** (`src/Inference.jl:50`): `ℓ(α) = loglikelihood(C(α), U)` becomes the weighted log-likelihood, and its Hessian is the observed information of the replicated sample. Bit-identical for unit weights, since `_weighted_loglikelihood` already is.
- **`:godambe` / `:godambe_pairwise`** (`:106`, `:120`) are a nonparametric bootstrap of the dependence measure, `idx[i] = rand(rng, 1:n)`. The bootstrap of the replicated sample draws `n` rows uniformly from it, that is observation `j` with probability `wⱼ/n`; a `StatsBase.sample!(rng, 1:n, fweights(w), idx)` and the measure on the resample unweighted. `n` stays `n`, so `Ω = n cov(M)` and `Vα = AΩA'/n` are untouched.
- **`:bootstrap`** (`:205`): the same resample, and `_refit` fits it **unweighted**, which means `_refit` strips `weights` from the recipe instead of refusing it (`_refit_kwargs` at `src/Fitting.jl:97` is where).
- **`:jackknife`** (`:209`): I would refuse it. The delete-one jackknife of the replicated sample removes one copy of observation `j`, which needs `wⱼ ≥ 1` for every `j`; after normalization to `n` that holds for unit weights only, and the delete-one-row jackknife with a leverage correction is the importance-weight construction that #516 did not settle. This is the one decision in here I would flag.

The composite GOF tests (`src/CopulaTest.jl:783`) keep refusing: the parametric bootstrap is the same construction, but the observed statistic needs a weighted empirical copula, which is a separate change.

## 3. The Sklar margins

`_estimate_sklar` refuses `copula_kwargs.weights` at `src/Fitting.jl:566` because its margins are fitted unweighted at `:574`, `Distributions.fit(D, x; margins_kwargs...)`. `Distributions.fit(D, x, w)` exists, `genericfit.jl:47` forwards it to `fit_mle(D, x, w)`, which goes through `suffstats(D, x, w)` and is defined for `Normal`, `Exponential`, `Gamma`, `Poisson`, `Geometric`, `Categorical` (and the multivariate families). So `weights` becomes a keyword of `fit(SklarDist{…}, X; weights)` itself, one vector read at every step:

- margin `i` by `Distributions.fit(Mᵢ, xᵢ, w; margins_kwargs...)`; a family without a weighted fit is refused by name (Distributions raises either a `MethodError` or its `"suffstats is not implemented"` `ErrorException`, so the refusal catches both);
- `:ecdf` (`:592`) by the weighted `pseudos`;
- the copula step with the same raw weights, which `_run_copula_estimator` normalizes itself so the margins, ranks, copula and recipe carry one vector;
- the stored log-likelihood (`:601`) by the weighted sum, so `aic`/`bic` follow and `nobs` (`:627`) stays `n`;
- the recipe carries the normalized weights, so `_copula_data`, `show` and `:bootstrap` read them.

Unit weights reproduce the unweighted margins up to rounding, not bit for bit: `suffstats(Normal, x, w)` reduces in another order than `mean`/`varm`. Integer weights reproduce the replicated fit to optimizer tolerance. `copula_kwargs.weights` still throws, pointing at the keyword. Note `fit(Beta, x)` and `fit(Cauchy, x)` are not `fit_mle`, so those margins are refused rather than silently changing estimator when weighted.

## Tests

- Kernels: unit weights bit-identical to `corkendall` / `corblomqvist` with and without ties and within an ulp of `corspearman`; integer weights equal the measure of the replicated sample; scale invariance; a zero weight removes its observation; a `NaN` column poisons its pair only.
- Every rank inversion (Clayton, Gumbel d=3, Gaussian d=2,3, Student `:itau` d=2,3 and `:itau_irho`, FGM, Galambos, Archimax, a survival copula, the independence copula): unit weights at three scales bit-identical; integer weights reproduce the replicated fit exactly, the Student profile to Brent's 1e-7.
- `:hessian` unit weights bit-identical, replicated to 1e-6; `:godambe` weighted vs replicated within Monte Carlo resolution at 3000 resamples; `:godambe_pairwise` in d=3; `:bootstrap` refits unweighted and `:mpl` stays `:mpl`; `:jackknife` refuses; the Sklar bootstrap runs.
- Sklar `:ifm` and `:ecdf` with `Normal`, `Exponential` margins: unit weights to 1e-10, replicated to 1e-6, `_copula_data` is the weighted `pseudos`, `fit(SklarDist)` and `fit(CopulaModel, …)` agree, a rank-inversion copula step works, `Beta` and `Cauchy` margins refused by name, `copula_kwargs.weights` refused.

## Cost

Nothing on the `weights === nothing` path. Weighted tau-b is O(n log n) with a Fenwick tree instead of the merge sort, about three times `corkendall` on n = 2·10⁵.

I have this implemented and tested on top of #532 and can send it as a stacked PR once you say the reading (a weight is a count) and the jackknife refusal are what you want.

dcelisgarza:
Stacked on #532: the diff shows both commits until #532 merges, review the last one. Closes #547.

## What

The three refusals #532 left, with the reading it gave a weight, "how many observations this column counts for", carried through: every procedure is the unweighted one on the sample in which observation `j` is repeated `weights[j]` times, written so that it extends to real weights and is invariant to their scale.

- **Rank inversions.** `:itau`, `:irho`, `:ibeta` and the Student `:itau_irho` invert the weighted sample measure. `_rank_measure(Val(:itau), U, weights)` returns `corkendall(U')` for `nothing` and the weighted kernel otherwise; every in-package rank `_fit` takes `weights=nothing` and calls it, the parameter-free families included. `:iupper` and the nonparametric estimators still refuse.
- **Inference.** `:hessian` inverts the observed information of the weighted log-likelihood. `:godambe`, `:godambe_pairwise` and `:bootstrap` draw each resample of size `n` with observation `j` taken with probability `weights[j]/n` (`StatsBase.sample!` with `fweights`), then compute the moment or refit the estimator on that resample **unweighted**; `_refit` strips `weights` from the recipe for that. `:jackknife` refuses a weighted model: the delete-one jackknife of the replicated sample needs every weight to be at least one, which after normalization to `n` holds for unit weights only. Composite GOF tests still refuse one.
- **Sklar route.** `fit(SklarDist{…}, X; weights)`: margin `i` by `Distributions.fit(Mᵢ, xᵢ, w)`, the weighted `fit_mle` through weighted `suffstats`, and a family without one refused by name (whether Distributions raises a `MethodError` or its `"suffstats is not implemented"` fallback); `:ecdf` by the weighted `pseudos`; the copula step with the same raw weights, normalized once there; the stored log-likelihood weighted; the recipe carries the normalized weights so `_copula_data`, `show` and `:bootstrap` read them. `copula_kwargs.weights` still throws, pointing at the keyword.

## How

Weighted Kendall's tau-b is one O(n log n) sweep over `x` with a Fenwick tree over the dense ranks of `y`, each `x`-tie block queried before it is inserted, weight sums in place of counts:

```
τ = Σ_{i<j} wᵢwⱼ sgn(xᵢ−xⱼ) sgn(yᵢ−yⱼ) / √((W₀−W₁)(W₀−W₂)),  W₀ = Σ_{i<j} wᵢwⱼ,  W₁ = Σ_g (T_g² − Σ_{i∈g} wᵢ²)/2
```

The pairs formed by the copies of one observation cancel between the two, so `W₀ − W₁ = (W² − Σ_g T_g²)/2` is the replicated sample's pair count, and the final `S / sqrt(a * b)` is `corkendall`'s arithmetic on exact integers, hence bit-identical for unit weights. Spearman is the weighted Pearson correlation of the weighted average ranks (`B + (T+1)/2` for a tie block of weight `T` after weight `B`, which is `tiedrank` for unit weights), a different reduction from `corspearman`'s, within an ulp. Blomqvist splits those ranks at the median rank `(W+1)/2` and counts weighted mass, bit-identical; the multivariate `β(U)` weights its orthant count.

## Not in this PR

- `:jackknife` on a weighted model, and the composite GOF tests on one (the observed statistic needs a weighted empirical copula).
- `:iupper` with weights.
- Importance weights, whose sandwich scales the scores by `wᵢ²`; nothing here decides what a weight *is* beyond what `fit` already said.

## Tests

- Kernels: unit weights bit-identical to `corkendall` and `corblomqvist` with and without ties and within an ulp of `corspearman`; integer weights equal the measure of the replicated sample; scale invariance; a zero weight removes its observation; a `NaN` column poisons its pair only; symmetry and a unit diagonal.
- Every rank inversion in Clayton, Gumbel (d=3), Gaussian (d=2, 3), Student `:itau` (d=2, 3) and `:itau_irho`, FGM, Galambos, Archimax, a survival copula and the independence copula: unit weights at three scales bit-identical to the unweighted `CopulaModel`; integer weights summing to `n` reproduce the replicated fit exactly (the Student profile to Brent's 1e-7).
- Inference: `:hessian` unit weights bit-identical and replicated to 1e-6 in three families; `:godambe` deterministic under a seeded `rng`, and within 10% of the replicated sample's at 3000 resamples; `:godambe_pairwise` in d=3; `:bootstrap` refits each resample unweighted and an `:mpl` fit stays `:mpl`; the Sklar bootstrap runs; `:jackknife` refuses.
- Sklar `:ifm` and `:ecdf` with `Normal`, `Exponential` margins: unit weights to 1e-10, replicated coefficients to 1e-6 and log-likelihood to 1e-8, `nobs` stays `n`, `_copula_data` is the weighted `pseudos`, `fit(SklarDist)` and `fit(CopulaModel, …)` agree, a rank-inversion copula step works, `show` reports the weights, `Beta` and `Cauchy` margins are refused by name, `copula_kwargs.weights` and negative weights are refused.

Locally (Julia 1.13, single thread): `Aqua.jl` (10), `api/utilities.jl` (127), `operations/fitting.jl` (2130), `operations/inference.jl` (55), `operations/fitting_selection.jl` (41), `operations/hypothesis_testing.jl` (173), `correctness/elliptical.jl` (30), `correctness/extreme_value.jl` (364), `correctness/bivariate_families.jl` (60), `correctness/archimedean.jl` (36), `correctness/nested_archimedean.jl` (85).

Cost: nothing on the `nothing` path; weighted tau-b is about three times `corkendall` on n = 2·10⁵.


dcelisgarza:
I'll bring it up to date.

dcelisgarza:
Wait, i just found i wasn't promoting some types.

dcelisgarza:
Rebased on `main` at 1d85de1, after #532 and #552 merged; the diff is now this PR alone. The zero-weight `_weighted_sample` pair and the single-LBFGS objectives from the #532 review were already carried here.

Two things in the last commit.

**The kernels' element type.** The weighted rank measures took their working type from the weights alone, so a `BigFloat` sample with `Float64` weights was ranked and summed in `Float64`. Each kernel now takes `promote_type` of its inputs' element types, as the weighted `pseudos` does after your note on #532, and the measure matrix is allocated in that type: `Float32`/`Float32` gives `Float32`, `Float32`/`Float64` gives `Float64`, `BigFloat`/`Float64` gives `BigFloat`, and `Float64` inputs are unchanged bit for bit. Tested on all three kernels.

**One test moved.** Under `:ifm` the margins are the copula step's input, and unit weights reproduce them to rounding only (Distributions.jl reduces its weighted sufficient statistics in another order). On the Clayton generator of #552 that ulp is enough to move the LBFGS trajectory: the unit-weight and unweighted Clayton θ̂ of the Sklar `:ifm` fit now land 2e-8 apart, which the old `rtol=1e-10` refused. The margins stay at 1e-10 and the copula parameter is compared at the optimizer's tolerance, as the replicated-sample fit already was. `:ecdf` is unaffected, since its copula step never sees the margins.

Locally (Julia 1.13, single thread): `operations/fitting.jl` (819 weighted tests) and `operations/inference.jl` (50) green on the rebase.

lrnv:
I went through the PR with my favorite LLM. There are two points I would change before merging, my llm will handle them dont worry.

### 1. `_fit_margin` catches `MethodError` too broadly

The current logic catches `Distributions.fit(D, x, w)` and translates a `MethodError` into “this margin has no weighted fit” based on `err.f`.

That can misdiagnose a genuine bug inside a weighted fitting implementation. For example, a valid `fit(D, x, w)` method could call another function internally which throws a `MethodError`; depending on how the call is wrapped, we may turn that implementation error into a misleading “weighted fit is unavailable” `ArgumentError`.

I would prefer checking applicability/capability before entering the fit, or otherwise making the exception test strict enough that we only catch the exact missing outer dispatch. Internal errors from an existing weighted fit should propagate unchanged.

### 2. `weights=ones(n)` should probably not disable jackknife

The PR explicitly establishes that uniform weights reproduce the ordinary unweighted estimator, and the rest of the implementation/tests make a strong effort to preserve that equivalence.

But `_infer(..., Val(:jackknife))` currently refuses any model whose recipe contains `weights`, including `weights=ones(n)`.

That makes these two mathematically identical fits behave differently:

```julia
fit(CopulaModel, C, U)
fit(CopulaModel, C, U; weights=ones(n))
```

The explanation in the error message itself says the replicated-sample delete-one jackknife is valid when every normalized weight is at least one, “which after normalization to `n` holds for unit weights only”. So the unit-weight case seems like it should simply use the ordinary jackknife.

More generally, I think the invariant worth preserving is: whenever the normalized weights are exactly all ones, all downstream behaviour should be indistinguishable from the unweighted fit, not only the coefficient estimates.

dcelisgarza:
Brilliant, thanks Oskar for taking all these requests. I may be back after we get VineCopulas sorted and I start implementing the prior.

PR #549, opened 2026-09-16
Document information-criterion semantics beyond MLE

### LINKED ISSUE #528: Define information-criterion semantics for non-likelihood and sequential fits before 1.0
## API/statistical contract to settle

`CopulaModel` currently exposes `StatsBase.aic`, `bic`, `aicc`, and `hqc` for every fitted model by evaluating the fitted distribution's log-likelihood and applying the usual parameter penalty. Automatic family selection likewise accepts any fitting `method` and ranks candidates by these quantities.

That is unproblematic for ordinary MLE. It is not the standard information-criterion setting for several other public estimators:

- `:itau`, `:irho`, `:ibeta`, etc. do not maximize the likelihood;
- MPL estimates from ranked data rather than an ordinary observed-data likelihood;
- `SklarDist` IFM is explicitly documented as sequential and **not** a joint maximum-likelihood estimate;
- `SklarDist` ECDF/rank fitting is likewise not joint likelihood maximization.

The manual already acknowledges part of this by saying:

> Use maximum-likelihood fitting for the usual information-criterion interpretation; passing another fitting method merely compares the scores at those estimates.

But the public API still calls the returned values AIC/BIC/AICc/HQC, and `fit(CopulaModel, Copula, U; candidates, criterion=:bic, method=:itau)` will automatically select a winner from them.

## Why decide now

This is observable 1.0 behavior and likely to be interpreted statistically, not merely computationally. Once users depend on `aic(rank_fitted_model)` or non-MLE automatic selection, tightening the contract later is a breaking change.

## Possible directions

Any of these could be defensible, but the policy should be explicit:

1. expose classical information criteria only when the estimator is a likelihood maximizer for the likelihood being scored, and reject other fits;
2. keep the numerical penalized-loglikelihood scores for non-MLE fits but expose them under different names / metadata and reserve `aic`/`bic` for their standard meaning;
3. deliberately retain the current generalized convention, but document it prominently in the StatsBase method docstrings and selection API, including IFM/MPL caveats and what comparisons are scientifically meaningful.

Automatic selection should follow the same policy instead of having a looser contract than direct `aic`/`bic` calls.

## Acceptance criteria

- The package states which estimator classes admit AIC/BIC/AICc/HQC under the 1.0 contract.
- `StatsBase.aic/bic` and automatic selection enforce or clearly encode that policy.
- Rank fits, MPL, Sklar IFM and Sklar ECDF are each addressed explicitly.
- Selection cannot silently present a non-classical score as ordinary BIC without the chosen contract making that intentional.
- Tests cover MLE and at least one non-likelihood estimator.

This issue does not require removing the cached fitted log-likelihood; that value is independently useful as a model diagnostic.

lrnv:
Closes #528.

## What

Documents the 1.0 decision to keep AIC, BIC, AICc, and HQC available for every `CopulaModel`, including MPL, rank-inversion fits, and sequential `SklarDist` fits.

The contract is operational: the package keeps using the existing penalized fitted-log-likelihood formulas for all estimators, while documenting that the usual classical information-criterion interpretation is not automatically justified outside ordinary MLE.

The documentation also points to adapted criteria from the literature (composite/pseudo-likelihood corrections and copula information criteria for two-stage/IFM estimation) as future additive work rather than changing the meaning of the existing 1.0 API.


lrnv (commented):
LGTM

PR #552, opened 2026-09-17
Write the Clayton generator in log1p/expm1 form

dcelisgarza:
## What

`ClaytonGenerator`'s power forms cancel below `eps`: for `θ < 2e-16`, `t^(-θ)` rounds to `1`, `ϕ⁻¹(t) = (t^(-θ) - 1)/θ` is exactly `0`, and every CDF value is `ϕ(0) = 1`:

```julia
julia> [cdf(ClaytonCopula(2, θ), [0.5, 0.5]) for θ in (1e-3, 1e-12, 1e-17, 1e-100)]
[0.2501200588787727, 0.24996196488587238, 1.0, 1.0]   # main; four digits already gone at 1e-12
julia> Copulas.β(ClaytonCopula(2, 1e-17))
3.0
```

The user-visible consequence is `:ibeta`: `_fit(::Type{<:ArchimedeanCopula}, U, ::Val{:ibeta})` brackets `Roots.find_zero` on `(-1 + 1e-8, 5)` for Clayton, bisection's bit-wise midpoint of a bracket across zero is `0.0` and then `2.4e-154`, and the false `β = 3` there swallows the true root, so **every Clayton `:ibeta` fit returned `θ = 0`**:

```julia
julia> U = rand(StableRNG(3), ClaytonCopula{2}(2.0), 200);
julia> Copulas.β(U), params(fit(ClaytonCopula, U; method=:ibeta)).θ
(0.41, 0.0)     # main
(0.41, 1.366)   # this PR
```

## How

`ϕ`, `ϕ⁻¹`, `ϕ⁽¹⁾`, `ϕ⁽ᵏ⁾` and `ϕ⁽ᵏ⁾⁻¹` in `log1p`/`expm1` form (`exp(-log1p(θt)/θ)`, `expm1(-θ log t)/θ`, …), which are exact where the power forms cancel and reach the `θ = 0` branch continuously; a non-finite `θ` keeps the power form and its limits. The specialized `_archimedean_logpdf` carried the same sum, `Σ tᵢ^(-θ) - d + 1`, and accumulates `Σ expm1(-θ log tᵢ)` with `log1p` on the last factor. The Clayton-over-Clayton `composition_taylor` is untouched: its base is `1 + θ_in t₀`, which does not cancel.

Every form takes its type from `θ` and `t` (Float32 and BigFloat are tested); the out-of-support branch of the derivatives returns that type's zero instead of the integer `0`.

## Tests

`correctness/archimedean.jl`: CDF, density and `β` at the independence values to 1e-11 for `θ ∈ {1e-12, 1e-17, 1e-100, 1e-300}`; agreement with the power forms to 1e-14 on `θ ∈ {-0.5, 0.3, 1.4, 2, 5}` inside the support, the `k`-th inverse round-tripping; the specialized log-density at `0` for `d = 2, 3, 4`; type propagation; `:ibeta` recovering `θ ≈ 1.37` on the sample above. The existing Clayton anchors, the nested Archimedean suite and the fitting suite pass locally on 1.13.

Found while extending #532 to the rank inversions: the weighted `:ibeta` reproduced the unweighted one bit for bit, both at `θ = 0`.


lrnv:
@dcelisgarza Same question: should i review this with red CI ? Please use draft mode until ready & green CI if you want me to wait, or tell me explicitely that you need my oppinon on something :)

dcelisgarza:
Sorry, i was looking into the failures. The aqua one is upstream. I had the same failure over in PortfolioOptimisers. I'll check what the benchmark is all about.

dcelisgarza:
Alright i found two sources of performance regression (julia was being clever with integer exponentiation, gone with the `log1p`  `expm1` change) and a `log` function being computed twice.

But the clanker found a `Core.Box` variable. ~I'll have it make the pr itself and rebase this and #532 onto it.~ I was mistaken, that was introduced in #532 so it'll be fixed as part of that pr.

Aaaand while it was at it it also figured out Aqua already has an update that fixes its failure.

dcelisgarza:
Three things in the red checks.

**Aqua.** The `Persistent tasks` error (`Unable to locate ChainRulesCore, a dependency of SpecialFunctions`) is Aqua 0.8.17, released 16:46 UTC today; `main`'s last green run resolved 0.8.16. Aqua 0.8.18 (its #400) fixes exactly this and is in General, so the next run resolves it without a change here. Whether to pin `Aqua` in the test compat is yours; I have not touched it.

**The `measure/clayton_*` regressions (+41% to +95%)** are the benchmark's θ = 2.0, not the generator. Julia's `Float64^Float64` takes a power-by-squaring fast path when the exponent is an integer, so on `main` `ϕ⁻¹(t) = (t^(-2.0) - 1)/2.0` costs ~6 ns while the `expm1(-θ log t)/θ` form costs ~20 ns at any θ. A fitted θ is never an integer, and at a non-integer θ the new forms are *faster* than the power forms (x86, scalar: ϕ 26 vs 32 ns, ϕ⁻¹ 20 vs 29 ns at θ = 2.1; 27 vs 54 ns at θ = 0.7). End to end, `measure/clayton_d5` on `main` is 702 µs at θ = 2.0 and 1454 µs at θ = 2.5; this branch is 1225 µs at both, 14% under `main` on the non-integer θ. The second commit moves the three measure fixtures to θ = 2.5 with a comment saying why. Tachometer will not gate this PR once `benchmark/` changed, per its README, so that is visible to you; if you would rather keep the benchmark as it is, drop that commit, and the alternative is an `isinteger(θ)` branch in `ϕ⁻¹` that keeps the power form for integer θ, which I did not want to add for a case no fit reaches.

**`fitting/clayton_mle_d2` +9% and `sklar_ifm` +8%** are the optimiser's trajectory: the d = 2 fit takes 105 loss evaluations on this branch against 87 on `main` (θ̂ moves at 1e-8), while d = 5 takes 168 against 165 and got 12% faster. Per call the new log-density was already faster (78 vs 100 ns at d = 2, 182 vs 239 ns at d = 5); the first commit takes `log(t)` once per coordinate instead of twice, which is bit-identical and brings it to 68 ns and 151 ns, and locally the d = 2, d = 5 and Sklar fits all sit under `main`. The `archimedean.jl` tests pass with both commits.

PR #554, opened 2026-09-17
Refactor fitting coordinates around Paramorph

lrnv:
## Summary

This PR refactors Copulas.jl fitting coordinates around `Paramorph.jl` and removes the legacy coordinate protocol rather than layering compatibility methods on top of it. The main goal is a smaller and more stable public API: model construction and natural parameters stay ordinary distribution concerns, while optimizer geometry lives in Paramorph.

### Architecture

- `Paramorph.param_space(CT, d)` is the single source of parameter geometry for generic copula fitting.
- The generic optimizer works in a flat vector of length `Paramorph.dimension(p)` and maps it through `Paramorph.constrain` directly into natural constructor parameters.
- `Paramorph.unconstrain` provides the inverse map used by inference and by initialization paths that already start from an existing natural parameter value.
- Archimedean copulas delegate fitting geometry to their generator; extreme-value copulas delegate it to their tail.
- `param_space` definitions are colocated with the corresponding distribution/family rather than living in the StatsBase coefficient adapter.
- `ParameterCoefficients.jl` is only the natural-parameter → `coef`/`coefnames` adapter plus the statistical-dimension → `dof` adapter. Natural coefficients are flattened mechanically and may contain constrained or redundant entries, while `dof` remains the intrinsic fitting dimension: normally the Paramorph dimension, or the explicit irreducible coordinate count for a custom runtime NAC reparameterization.
- Empirical plug-in objects explicitly use the zero-dimensional space `()` so fitted state is not mistaken for statistical parameters.
- Family-specific fitting algorithms remain only where the statistical/numerical algorithm or feasible set is genuinely specialized (for example correlation-factor likelihoods, Student profiling, empirical estimators, or the multivariate FGM polytope).
- The previous compatibility layer and the `_example`, `_unbound_params`, `_rebound_params`, `_construct_fitted_copula` protocol are removed rather than wrapped.

### Nested Archimedean copulas

`NestedArchimedeanCopula` construction remains intentionally permissive with respect to parent-child nesting theory, matching the pre-refactor API: the constructor validates tree structure, dimensions, coordinate placement and each node generator's own dimensional validity, but it does not try to certify the mathematical nesting relation.

Template fitting is stricter. `Paramorph.param_space(template)` builds the dependent parameter geometry that the generic fitter currently knows how to optimize. Finite unconstrained coordinates are mapped by `Paramorph.constrain` into that supported fitting region; an initial template outside it, or a tree with no implemented fitting geometry, is rejected before optimization. Unsupported template fitting therefore does not make explicit construction invalid.

This PR deliberately does not attempt to turn the package into a catalogue or theorem prover for all NAC nesting conditions. Additional fitting support should be expressed as Paramorph geometry plus focused regression tests, without expanding constructor-level theory.

### Public parameter API

`Distributions.params` remains a natural model representation in canonical constructor order. Paramorph is not an additional requirement for non-fitting public operations: a downstream generator that implements the documented minimal generator contract can still be constructed and queried through the distribution API without defining a fitting parameter space.

### StatsBase representation

`StatsBase.coef` is intentionally a representation of the fitted model's natural parameters, not a minimal coordinate chart. For example, a symmetric/correlation matrix is exposed in full, and `vcov` is the covariance on that full displayed representation. Such covariance matrices can legitimately be singular/positive-semidefinite because displayed coefficients may be fixed or redundant. `StatsBase.dof` remains the intrinsic fitting dimension and therefore need not equal `length(coef(model))`: it comes from Paramorph for ordinary parameter-space fits, while an explicitly supplied runtime NAC reparameterization uses its irreducible coordinate count.

### Performance

The fitting hot path does not build a `NamedTuple` and immediately call `values` on it. It uses `Paramorph.constrain` directly; names remain metadata for public parameter representations and diagnostics. Specialized numerical kernels are kept when they materially outperform the generic route; reusable coordinate transforms belong in Paramorph rather than being duplicated in Copulas.jl.

### Dependency note

`Paramorph.jl` is currently sourced directly from its `main` branch through `[sources]`, so this draft PR follows the current Paramorph development state. This PR must remain unmerged until Paramorph is registered in General.


lrnv:
Superseded by #561, which now targets main and contains the complete change set.

PR #557, opened 2026-09-19
Fix conditioning for bivariate Tawn copulas

### LINKED ISSUE #556: Condition fails for bivariate TawnCopula despite AsymLog equivalence
Hi @lrnv,

while migrating the extreme-value pair-copula code in VineCopulas.jl to rely only on the public `Copulas.jl` API, I noticed a small inconsistency with `TawnCopula`.

Currently `TawnCopula` can be constructed for `d = 2`, and Copulas.jl already tests the equivalence

```julia
α, θ1, θ2 = 2.1, 0.67, 0.38

A = ExtremeValueCopula{2}(AsymLogTail(α, θ1, θ2))
T = ExtremeValueCopula{2}(TawnTail(α, [θ2, θ1]))
```

so in the bivariate case this Tawn parameterization reduces to the historical asymmetric logistic model.

However, the public conditioning path does not currently preserve this equivalence. For example,

```julia
C = ExtremeValueCopula{2}(TawnTail(2.1, [0.38, 0.67]))
D = condition(C, 2, 0.7)
cdf(D, 0.4)
```

fails, while the equivalent `AsymLogTail` copula works.

From what I can see, `condition(::ExtremeValueCopula{2}, ...)` routes to `BivEVDistortion`, which expects the scalar bivariate Pickands interface `A`, `dA`, `d²A`. `TawnTail` is instead implemented through the general multivariate STDF `ℓ`, so this path is missing for `d = 2`.

I found this because I initially tried to add `TawnTail` to the VineCopulas pair benchmark, but I think this should probably be solved upstream rather than introducing a special workaround in VineCopulas.

Maybe for `d = 2` we could either provide the corresponding Pickands path for `TawnTail`, or route the bivariate Tawn representation through its equivalent asymmetric-logistic conditioning machinery.

What do you think?

Santymax98:
Fixes #556.

Bivariate `TawnTail` conditioning was incorrectly routed through the generic `BivEVDistortion` fast path, which requires the scalar Pickands interface.

This PR:

* restricts that fast path to `BivariatePickandsTail`;
* uses the exact bivariate Tawn → AsymLog equivalence for conditioning;
* preserves endpoint behavior and existing Pickands fast paths;
* adds regression tests for conditional CDF, density, quantiles, and endpoints.

No fitting or public API changes.


lrnv:
Very good job !

